PR #46, opened 2018-12-10
Remove spurious use of logos_derive

mikolajpp:
On nightly these additional uses end with error
```
error[E0252]: the name `Logos` is defined multiple times
 --> tests/tests/advanced.rs:2:5
  |
1 | use logos::{Logos, lookup};
  |             ----- previous import of the macro `Logos` here
2 | use logos_derive::Logos;
  |     ^^^^^^^^^^^^^^^^^^^ `Logos` reimported here
  |
  = note: `Logos` must be defined only once in the macro namespace of this module
help: you can use `as` to change the binding name of the import
```

### DIFF
--- a/tests/tests/advanced.rs
+++ b/tests/tests/advanced.rs
@@ -1,5 +1,4 @@
 use logos::{Logos, lookup};
-use logos_derive::Logos;
 use std::ops::Range;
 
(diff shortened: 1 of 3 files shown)


mikolajpp:
Hm, this is strange, because I remember being able to run tests on nightly before. 
Perhaps running them on stable is more preferable though.

PR #49, opened 2018-12-10
Another fork corner case

mikolajpp:
This is another corner case.

I think if you want to take time and redesign branching algorithm from scratch then it is fine by me 
to wait longer with the fix.  However it seems like fixing this case should already prevent a large 
class of future bugs, as this is pretty much the last bug standing in my lexer, so hotfix is also worth considering :-)

### DIFF
--- a/tests/tests/advanced.rs
+++ b/tests/tests/advanced.rs
@@ -40,6 +40,9 @@ enum Token {
     #[regex="~[mhs][0-9]+"]
     LiteralRelDate,
 
+    #[regex="~[0-9]+-?[\\.0-9a-f]+"]
+    LiteralAbsDate, 
+
     #[regex = "🦀+"]
     Rustaceans,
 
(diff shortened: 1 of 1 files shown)


maciejhirsz:
Tests are passing, so just adding the definition doesn't seem to break anything. Does it match incorrectly?

mikolajpp:
Ah, sorry about that. I assumed the ordering did not matter, when it is in fact ordering which triggers the bug again. Already pushed breaking commit.

mikolajpp:
Just to be explicit, because I put some junk into first commit.

This is double plus good. 
```
 #[regex="~[mhs][0-9]+"]
  LiteralRelDate,
  #[regex="~[0-9]+-?[\\.0-9a-f]+"]
  LiteralAbsDate, 
```

However reversing the variants breaks again.

maciejhirsz:
Fix released as #0.9.3 :).

PR #51, opened 2018-12-10
Quote fork bug

mikolajpp:
I am quite sure I have seen this one before. 
The example is with quotes, but it should work with any character. 
In short, having tokens for `'` and `'''` makes it an error to parse `''` (double single quote) 
instead of returning two tokens corresponding to `'`.

### DIFF
--- a/tests/tests/advanced.rs
+++ b/tests/tests/advanced.rs
@@ -43,6 +43,12 @@ enum Token {
     #[regex="~[mhs][0-9]+"]
     LiteralRelDate,
 
+    #[token="'"]
+    SingleQuote,
+
+    #[token="'''"]
+    TripleQuote, 
+
     #[regex = "🦀+"]
     Rustaceans,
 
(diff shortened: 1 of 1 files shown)


mikolajpp:
I am not very familiar with the branching machinery, but it seems logos is not backtracking after seeing two single quotes, expecting the third one because of `'''` token.

maciejhirsz:
Oh, yeah, this is a tricky one because no-backtracking is _kind of_ a feature. I've the same issue in Ratel since JS has `.` and `...` tokens, but no `..`.

I'm thinking it should be possible to speed things up a notch by having the `Source` read a fixed number of bytes into an array (`[u8; 2]` in this case) and then compare that to predefined arrays. It might also be a performance win since it could optimize two bounds checks and byte checks into single bounds check and one `u16` check.

mikolajpp:
Ah right! Totally forgot about this. I guess this is in the future; But the fixed width lookup sounds quite promising. 
For now I am just using a workaround in the form of "EmptyString" token to catch the double quote case. 

We can close this one as well and perhaps open a feature ticket.

maciejhirsz:
Made the issue, I'll merge the PR into another branch once I start working on this.

Issue #61, opened 2018-12-15
Sublexers using callbacks in documentation


mikolajpp:
In the end I am using the following construct for dispatching sublexer.

```
fn sublexer<'source, Src: Source<'source>, Subtok>(skip: usize, lex: &mut Lexer<Token, Src>) 
-> Vec<Subtok>
where Subtok: Logos + PartialEq + Clone  + fmt::Debug
{

    let range = (lex.range().start + skip)..lex.source.len();
    let slice = lex.source.slice(range).unwrap();

    let mut sublex = Subtok::lexer(str::from_utf8(&slice.as_bytes()).unwrap());
    let mut tokens = vec![];

    sublex.advance();

    println!("Sublexing...");

    while sublex.token != Subtok::END {

        println!("Got {:?} for {}", sublex.token, sublex.slice());
        if sublex.token == Subtok::ERROR {
            return tokens
        }

        tokens.push(sublex.token.clone());
        sublex.advance();
    }
    tokens
}
```

If you think it is worth it, we might want to put some short writeup in the docs about 
this.

mikolajpp:
This does not handle bumping top-level lexer, but it is easy to do that tracking farthest end of current sublexer token.

maciejhirsz:
I think it might be reasonable to add some API support for this to be less hacky.

mikolajpp:
Yes, for sure. I was thinking about hacking 
```
Intricate(ComplexToken)
```
into Logos, so that Intricate would fire ComplexToken lexer,  but recently I am short on time for getting enough rapport with the codebase. This would require some substantial changes, as for now Token variants are supposed to be strictly without fields. 

And from cursory thinking about this it is not clear for me what solution is the best. Should we have sublexers 
parse as long as they can? This is the way I have it working now, since some sublexers do not have natural ending token (think some embedded templating language). Alternatively, the parsing can be done just using advance as for top level lexer. A third way would be to have an API return ready made sublexers to be handled by the user. 

Once my lexer settles down a bit I will compile a list of things that I needed hacks to get done, and perhaps that could give us some useful pointers, as in terms of complexity and quirks I believe it is testing Logos quite extensively as indicated by the number of bugs smoked out ;-)

maciejhirsz:
I was thinking something in the lines of:

```
let sublexer = lexer.context_switch::<SubToken>();
// or
let sublexer = SubToken::sublexer(lexer);
```

[...]

Issue #27, opened 2018-11-28
Switching lexers


ronniec95:
Is there a way to switch lexers?
Standard case of lexing a buffer using a primary lexer and then a separate function to parse quoted strings, then switch back?

Is there an easy way to get the current position of the lexer and move it along while I take over?

maciejhirsz:
_Maybe_. There is no way to do context switching on the `Lexer` atm, but you can do a lot with function callbacks. Have a look at [this piece of code from Lunarity](https://github.com/paritytech/lunarity/blob/39fb996eab248a0a4888b90748a7c8198a0722f9/lexer/src/token.rs#L419-L438), it handleds the block comments (`/* ... */`) by exposing the internals of the `Lexer` and basically doing the work manually.

On a sidenote, quoted strings [you can already do with regex](https://github.com/maciejhirsz/logos/blob/cd6691715d8f597b4a1956734e73a38a673a0cf8/tests/tests/advanced.rs#L16-L17). If you want the lexer to produce an unescaped `String` out of the literal, you can add a `#[callback]` there that reads through it and stores the result on some fields on [`Extras`](https://github.com/maciejhirsz/logos/blob/cd6691715d8f597b4a1956734e73a38a673a0cf8/tests/tests/simple.rs#L8-L30).

jeertmans:
Closing this a too old, feel free to re-open if needed.

PR #65, opened 2018-12-17
Mighty Morphin Logos Lexer

CAD97:
Relevant to #61, #27. Probably closes one or both of them.

Adds the capability for lexers to change the token type that they're lexing.

### DIFF
--- a/logos/src/lexer.rs
+++ b/logos/src/lexer.rs
@@ -11,6 +11,7 @@ pub type Lexicon<Lexer> = [Option<fn(&mut Lexer)>; 256];
 
 /// `Lexer` is the main struct of the crate that allows you to read through a
 /// `Source` and produce tokens for enums implementing the `Logos` trait.
+#[derive(Clone)]
 pub struct Lexer<Token: Logos, Source> {
     /// Source from which the Lexer is reading tokens.
     pub source: Source,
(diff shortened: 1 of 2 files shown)


maciejhirsz:
Not sure about `morph` as a name, but we can bike-shed later :sweat_smile:.

Considering the differences between `morph` and `morph_extras`:
+ If `Extras` on either lexer are of type `()` then there is no drawback to always using `morph_extras`.
+ If `Extras` on parent is of some type `T`, but on the new lexer it is `U`, then we can require that `T: Into<U>` thus easily convert between `Extras` for given lexer modes.
+ We can provide a generic `impl<E: Extras> From<E> for ()` (if Rust complains, can just create custom 0-sized `struct NoExtras` instead) so that if parent has extras of some type `T`,but new lexer is not defined, then the extras are dropped.

With the points above, I think we could just have a single `morph` method that satisfies all use cases.

## Randomness

Would be cool if the method was a `const fn`, though that won't work with `Into`, and although we could provide our own trait for that (`ConstInto`? :confused:). It's likely to get inlined anyway.

## Bike-shed

I was considering having a `context_switch` method, though if we are switching _modes_ (as pert the [article](https://www.oilshell.org/blog/2017/12/17.html)) then `switch` would have been sufficient, _but_ that might be confusing to people who expect `switch` to be something like `match` (C/Java baggage)?

CAD97:
> Could move `extras` field on the struct to be the last position, so that if it changes across lexers, we don't have to shuffle `token_start` or `token_end` around in memory (I reckon LLVM will be smart enough here to do as little copying as possible). This would be a super micro-optimization that probably doesn't matter at all in any meaningful way, but eh.

Field order on-disk is already undefined on `#[repr(Rust)]`, so changing the order of fields (shouldn't) change anything. For that matter, I _really_ don't expect swapping modes with changing extras to be a hot path. Even then, I doubt we can do much (if anything) to improve the result LLVM produces here. Ideally, morphing between two same-extras token types should be a no-op.

> differences between `morph` and `morph_extras`

Unfortunately, you can't have `impl<T: Trait> From<T> for Concrete` and `impl Trait for Concrete` (without specialization) because of the reflexive `impl<T> From<T> for T`. However, using `Into` for the conversion seems reasonable, and users can impl `From` if they need to change between extras.

> actually, if we use a single method, couldn't it be just `impl<T, U, S> From<Lexer<T, S>> for Lexer<U, S> where ...`?

I think using a dedicated function is more useful, not least because it has non-trivial semantics to the state the new lexer is put in (even if those semantics are effectively a no-op). Additionally, I think the `advance_as` is definitely _very_ useful and evocative; it's `advance` but with a new token type. I'd be fine just exposing `advance_as`, but I think exposing either `morph` or `advance_as` as `From::from` is a footgun waiting to happen.

> bikeshed

`morph` was chosen as a name mainly so I could make the Mighty Morphin Power Rangers joke. I can leave that behind if we can come up with a better name. `Into::into` seems like a good name, but I'm wary of hiding the semantic complexity behind such a common name. If we keep `morph` rather than just exposing `advance_as`, I'm not sure what the best name would be. Hence picking something flavorful instead of semantic.

I think `advance_as` should be the main method people use though, being its semantics are "`advance` with a new token type". For that reason `advance_as::<Token>` feels like the correct name, but again, I could go either way.

CAD97:
I've added the test example for the lexer mode morphing. An interesting point to make is that with the `Extras` bound, Rust isn't able to determine the token type without having it hinted, so a turbofish or type bound for the lexer is required.

PR #93, opened 2020-03-21
runnable `cargo test`

k-nasa:
Now, if i run `cargo run`, it failed.
I fixed it.

### DIFF
--- a/tests/tests/lexer_modes.rs
+++ b/tests/tests/lexer_modes.rs
@@ -1,8 +1,7 @@
-use logos::Logos;
+use logos::Logos as _;
 use logos_derive::Logos;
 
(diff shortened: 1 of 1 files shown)


tests/tests/lexer_modes.rs
```diff
@@ -1,8 +1,7 @@
-use logos::Logos;
+use logos::Logos as _;
```
maciejhirsz: This doesn't look right. The `logos::Logos` is a trait, while the `logos_derive::Logos` is a derive macro, they _can_ and _should_ coexist in the same namespace.

maciejhirsz: Heh, you are right, weird as this is...

Issue #88, opened 2019-10-28
Some kind of regression between 0.9.7 and 0.10.


alexschrod:
To be specific, I'm getting this issue with 7a31ef303a2d90cbf91ccd8cf20931ac020e9cc2, but not with efc666dd6e8f7ab598a1d1d6850fe3b1e564f77a. (The ones in-between causes me compilation errors, so it's hard to say which of them introduced the code that leads to this problem.)

This is my configuration:

<details>
<summary>Token type</summary>

```rust
#[derive(Debug, PartialEq, Clone, Copy, Logos)]
pub enum Token {
    #[end]
    EndOfProgram,

    #[regex = "\"[^\"]*\""]
    Text,

    #[regex = "'(\\\\x[0-9a-fA-F]+|\\\\[0-7]+|\\\\(n|t|r|\\\\|\"|'))'"]
    Char,

    #[regex = "-?([0-9]*\\.[0-9]+|[0-9]+\\.[0-9]*)"]
    Float,

    #[regex = "-?0(x|X)[0-9a-fA-F]+"]
    Hex,

    #[regex = "-?[0-9]+"]
    Integer,

    #[regex = "(struct|bool|char|int|float|string|color)_t"]
    Type,

    #[regex = "[a-zA-Z_]+"]
    Identifier,

    #[regex = "#[0-9a-fA-F][0-9a-fA-F][0-9a-fA-F][0-9a-fA-F][0-9a-fA-F][0-9a-fA-F]"]
    Color,

    #[token = "{"]
    BraceOpen,

    #[token = "}"]
    BraceClose,

    #[token = "="]
    Assign,

    #[token = ","]
    Comma,

    #[token = "["]
    BracketOpen,

    #[token = "]"]
    BracketClose,

    #[regex = "//[^\n]*"]
    #[token = "/*"]
    #[callback = "ignore_comments"]
    #[error]
    UnexpectedToken,
    UnexpectedEndOfProgram,
}
```
</details>
<details>
<summary>Sufficient amount of document causing lexing issues in 0.10</summary>

```
tileset "Default Tileset" {
	tileWidth=16
	tileHeight=16
	description="Foobar"
	author="BazQuux"
	version="0.15"

	texture "terrain.png"
	{
		terrain_sprite {
			sprites=[0]
		}

		terrain_sprite "grass" {
			sprites=[
			// Height < -1.0
				20,
			// Height < -0.5
				21,
			// Height < 0.0
				22,
			// Height < 0.25
[...]

Issue #87, opened 2019-10-10
The macro #[regex=r"[\w]+"] make rustc get stuck and crash.


pzzp:
![image](https://user-images.githubusercontent.com/21138299/66552091-515c0280-eb7b-11e9-8dcf-bd1de10a7790.png)

Look at the picture. 
If I didn't cancel it, the rustc will be stuck for a long time and then tell a error.  Sometimes it even make my system crash.

maciejhirsz:
This should be fixed in 0.10.0-rc3.

Issue #81, opened 2019-06-24
Incorrect regex with parentheses


wirelyre:
This lexer:

```rust
#[derive(Logos, Debug, PartialEq)]
enum Token {
    #[end] End,
    #[error] Error,

    #[regex = "a(a)+"] Aaa,
}
```

incorrectly matches a single `a`. Changing the regex to `aa+` makes it work correctly.

maciejhirsz:
This should be fixed with 0.10.0-rc3, please re-open if it happens again.

Issue #80, opened 2019-06-21
logos_derive stack overflow in regex macro


2over12:
Not sure if this is related to the previous issues with the regex macro but this code results in stack overflow in the compiler.

```rust
extern crate logos;

use logos::Logos;

#[derive(Logos, Debug, PartialEq)]
enum Token {
    #[end]
    End,
    #[error]
    Error,
    #[regex = "[0-9]*.?[0-9]+"]
    Text,
}

fn main() {}
```

maciejhirsz:
Aggregate into #70.

PR #94, opened 2020-03-22
Refactor logos-derive from tree to graph

maciejhirsz:
Progress tracking:

+ [x] new `Graph` primitives
+ [x] parsing `#[token]` declarations to `Graph`
+ [x] parsing `#[regex]` declarations to `Graph`
  + [x] byte concatenations
  + [x] zero-or-one (`?`) groups
  + [x] zero-or-more (`*`) groups
  + [x] one-or-more (`+`) groups
  + [x] alternations (`a|b`)
  + [x] group concatenation
  + [x] reimplement full unicode handling with `utf8-ranges`
+ [x] merging non-looping branches (`foo` with `foobar`)
+ [x] merging looping branches (`[0-9]+` with `[0-9]+\.[0-9]+`)
+ [x] merging looping branches with non-looping branches (`[a-z]+` with `foobar`)
+ [x] token disambiguation
+ [x] handle trivia
+ [x] code generation, remove `lexicon` from the `Logos` trait
+ [x] code optimization: parse multiple bytes at a time
+ [x] callbacks and extras
+ [x] code optimization: turn complex forks into jump tables

---

I now believe that most of the issues current implementation has (#87, #81, #80, #79, #78, #70, and probably more) is due to the fact that trying to construct a tree is just not the right way of even trying to solve the problem.

What I think is a solution is a complete rewrite of the `logos-derive` from tree to a graph that can more adequately represent loops and arbitrary state jumps without the explosive nature of trying to build up all possible permutations in a tree. All the nodes of the graph are going to be stored on a single `Vec` based struct (called `Graph`), and will be referenced by their index in that `Vec`. The nodes are going to be immutable, so any permutations (merging forks) will have to create a new node with a new id.

Here is a current (custom) debug print for what I imagine a simple `[a-z]*` regex should look like in the graph:
```
[
    :0 "IDENT",
    :1 [
        [a-z] ⇒ :1,
        ⤷ :0,
    ],
]
```

The node `:0` is a token, the node `:1` is a fork with a single arm matching one byte to range `[a-z]`, on success we navigate to node `:1` (creating a loop), on miss we navigate to node `:0` and return a token. Generating Rust code out of this should be pretty straight forward, we can make every node a function definition, every jump to a function call (loops shouldn't lose performance due to tail call recursion). There is going to be room for optimization in code generation, although LLVM is probably going to do a better job at figuring out how and when to inline stuff than I ever will.

This removes the need of marking forks as `Plain | Maybe | Repeat`, and it should also remove the need for the `fallback` on branches, that really was just a hack to make identifiers working alongside named keywords.

Going to leave this draft open for comments (CC #88).


maciejhirsz:
While I'm at it, derive will now report error more than one error at once, and will do so spanned to where the problem is:

```
error: `Token::End` has a discriminant value set. This is not allowed for Tokens.
 --> tests/tests/dev.rs:8:5
  |
8 | /     #[end]
9 | |     End = 1,
  | |___________^

error: Only one #[error] variant can be declared.
  --> tests/tests/dev.rs:14:5
   |
14 | /     #[token = "hello!"]
15 | |     #[error]
16 | |     Hello,
   | |_________^

error: Previously declared #[error]:
 --> tests/tests/dev.rs:5:5
  |
5 | /     #[error]
6 | |     Error,
  | |_________^

error: aborting due to 3 previous errors
```

wirelyre:
I also started exploring this in wirelyre/logos@8db611da674deccdf9dbb2df4d37cc4d0f16a6d8.

My design was to not rely on tail recursion by using labeled loops:

```rust
'n0: loop { match src.next() {
    'a' => 'n1: loop { match src.next() {
        'a' => continue 'n1,
        'b' => continue 'n0,
    }}
    'c' => 'n2: loop { /* etc. */ }
}
```

This works whenever nodes are nested (no jumps from `'n1` to `'n2`), which in practice is almost always. In other cases you have to duplicate part of the graph inside the loops.

I have some notes on how well LLVM tends to optimize that; I'm not sure how well it will inline the tail recursive nodes but it's bound to be less predictable.

I found it useful to:
- Remove `Regex` and match only one character at a time
- [Replace `Pattern`](https://github.com/wirelyre/logos/commit/9056bda09d1a4db82fe5526b1f16788cad2e4678) with a bitset

Contrary to the code I wrote, I planned to:
- Construct an NFA from all patterns at once, then generate the graph directly from that NFA
- Remove the graph library since the algorithms aren't useful

wirelyre:
Also I was going to use this disambiguation strategy:

1. Take the longest tokens or regexes that match.
2. If both a token and regex match, return the token.
3. If two regexes match, return the first one. (But show a warning at compile time.)

This correctly deals with identifiers and keywords (`let` vs. `letter`); and integers and floats (`123` vs. `123.45`).

maciejhirsz:
>     * Remove `Regex` and match only one character at a time
/9056bda09d1a4db82fe5526b1f16788cad2e4678) with a bitset

There is an optimization in place where for any branch, and any fork with branches that lave length > 1, you can read multiple bytes at once and avoid doing bounds checking. This is especially useful when a branch is a byte sequence of 4 or 8 bytes, you can load [u8; 4] or [u8; 8], since LLVM can optimize those compares to 32/64 bit integer.

>     * [Replace `Pattern`](https://github.com/wirelyre/logos/commit/9056bda09d1a4db82fe5526b1f16788cad2e4678) with a bitset

Pattern is definitely a mess. In this rewrite I've already opted not to use it. All my sequences are just made of `Range(u8, u8)`, while the fork is using a table 256 ids long, so I just splatter the ranges onto that and never have to worry about having to find where the sets are overlapping.

wirelyre:
> you can read multiple bytes at once

Yes, sorry, I meant only "matching" single bytes in the graph, not actually in the generated code.

That seemed promising because it simplified the graph structure. And then any optimizations could logically be done on the graph as a complete data structure, rather than making the edges more complex, if that makes sense.

For instance this would seem to leave room for:

```
tokens "aaaa", "aabb"; regex "[a-z]+"
```

making

```rust
match chunk_of_4 {
    b"aaaa" => _,
    b"aabb" => _,
    _ => { if regex(b[0]) && regex(b[1]) /* ... */ }
}
```

because even though all three patterns share the first three nodes, so they'll be in separate `Branch`es, if we have the whole graph we can recognize the larger pattern.

Although writing this it's now clear that loops are a big tradeoff, because in this case you'd need to duplicate the regex state like eight times.

Evrey:
Hey there. =) Just lemme interlude for a few words:

I'm glad that `logos` is visibly back in the works and I'm highly excited for this PR. Here, have some love: ♥️

maciejhirsz:
@Evrey thanks!

So with the new graph a single `\w` is creating a total of 278 nodes, which is not bad considering that there is a total of 883 possible sequences of `utf8_ranges` that this is constructed from. If I reintroduced `Pattern` or `Class` as it was done before, it could probably produce fewer nodes (right now what would be a single byte `Pattern` that can span multiple ranges has to be a fork). I think I'll carry on with what I have right now as the final Rust code will need to expand those to either lookup tables or `match`es anyway.

Now, the really good news is that `\w\w` produces 555 nodes, which is, as expected, a double (-1 for the root note). Current master produces [228k LOC](https://github.com/maciejhirsz/logos/issues/70#issuecomment-470269926) of Rust for that input since the tree explodes when one whole tree of `\w` gets another whole tree of `\w` attached to _every leaf_ (of which I counted 303 in the graph printout there). I'm really happy with my progress on this so far, and it look like after a year the CPU melting during compilation is going to be put to rest.

maciejhirsz:
Okay, 5 days in I think we are pretty much on schedule of my initial estimate of 2 weeks for this. Bar bugs and random tweaks to improve the denseness of the graph, I reckon the graph construction part of the equation is done. The following token definitions:
```
    #[token = "foobar"]
    Foobar,

    #[regex = r"[a-z]+"]
    Identifier,
```

Produce the following debug print of the graph (node 9 is the entry):
```
{
    2: ::Foobar,
    3: ::Identifier,
    5: {
        [a-z] ⇒ 5,
        _ ⇒ 3,
    },
    7: [
        oobar ⇒ 8,
        _ ⇒ 5*,
    ],
    8: {
        [a-z] ⇒ 5,
        _ ⇒ 2,
    },
    9: {
        [a-e] ⇒ 5,
        f ⇒ 7,
        [g-z] ⇒ 5,
    },
}
```

This adequately maps all the regex to something that should be fairly straight forward to write code generation for. There are no fallback hacks, no loop designation, everything is dead simple. The only quirk is that ropes (old standalone branches) can have miss values annotated as `Miss::None`, `Miss::First(NodeId)`, or `Miss::Any(NodeId)`. Node 7 in example above has `Miss::Any(5)` (marked with `*`), which simply means that any partial parse, such as `(f)oo`, ought to jump to node 5 instead of error.

The whole `graph` module in generic over `Leaf`, and the only dependencies inside it are of `regex_syntax`, `utf8_ranges`, `fnv`, and the `Error` type from the crate. Once this whole thing is running it should be easy to move this whole module to it's own crate, closing #20.

The only trait bound required for the generic `Leaf` is the `Disambiguate` trait, which is almost identical to `std::cmp::Ord`, and is used to disambiguate token conflicts. Current disambiguation strategy is to give each token node a priority equal to it's length not counting repetition, so that:

+ `[a-z]+` has priority of 1.
+ `foobar` has priority of 6.
+ `(fooz|bar)+qux` has priority of 7 (`foozqux` is the longest product without repeats).

Since the priority is recorded on the token node, we could in the future add an attribute that allows for setting an arbitrary priority to a token, if some finer control is needed.

maciejhirsz:
This test doesn't run trivia, which isn't in yet. My really crude code generator managed to produce something reasonable for it.

```
running 1 test
test binary::handles_non_utf8 ... ok
```

maciejhirsz:
Tests with trivia stopping here:
```
running 12 tests
test advanced::cyrillic ... ok
test advanced::lookup ... ok
test advanced::integer ... FAILED
test advanced::float ... FAILED
test advanced::keywords ... FAILED
test advanced::rustaceans ... ok
test advanced::hex ... FAILED
test advanced::polish ... ok
test advanced::sigs ... ok
test advanced::subquotes ... ok
test advanced::what_the_heck ... ok
test advanced::string ... FAILED
```
All or most of the failed tests are because lookahead is not done yet in generated code.

Performance with barely any optimizations performed on the generated code:
```
test identifiers                       ... bench:       1,328 ns/iter (+/- 121) = 586 MB/s
test keywords_operators_and_punctators ... bench:       3,036 ns/iter (+/- 110) = 701 MB/s
```

maciejhirsz:
Getting closer to numbers on `master` :)
```
running 2 tests
test identifiers                       ... bench:       1,026 ns/iter (+/- 66) = 759 MB/s
test keywords_operators_and_punctators ... bench:       2,161 ns/iter (+/- 82) = 986 MB/s
```

maciejhirsz:
Progress

```
running 12 tests
test advanced::float ... ok
test advanced::cyrillic ... ok
test advanced::hex ... ok
test advanced::integer ... ok
test advanced::keywords ... ok
test advanced::polish ... ok
test advanced::rustaceans ... ok
test advanced::lookup ... ok
test advanced::sigs ... ok
test advanced::what_the_heck ... ok
test advanced::subquotes ... ok
test advanced::string ... FAILED
```

maciejhirsz:
All tests but 3 are :heavy_check_mark:. 2 are failing because callbacks are not implemented, one is what looks like a disambiguation weirdness that I need to take a better look at.

Hopefully I can fix those quickly tomorrow, and then start having fun with optimizations :D.

maciejhirsz:
Debugging took a while, but we [are all green :heavy_check_mark:](https://travis-ci.org/github/maciejhirsz/logos/builds/668480714). There might still be some edge cases, turns out the graph makes everything simpler and easier, but it's hard to produce a canonical structure for every possible permutation of regex.

Unwinding loops and simple lookup tables for multiple ranges, and numbers are getting there:
```
test identifiers                       ... bench:         813 ns/iter (+/- 56) = 958 MB/s
test keywords_operators_and_punctators ... bench:       2,563 ns/iter (+/- 105) = 831 MB/s
```
Next I need to do jump tables for expensive match branches, and more aggressively read multiple bytes at a time when possible, and then I reckon this thing is home.

maciejhirsz:
Optimized branching :)

```
test identifiers                       ... bench:         681 ns/iter (+/- 54) = 1143 MB/s
test keywords_operators_and_punctators ... bench:       2,315 ns/iter (+/- 162) = 920 MB/s
```

ratmice:
So, tried this out on a branch of a project of mine, 

I had to fix a bug in my lexer (mystery solved!), but it went green,
after fixing that bug though 0.10-rc2 which was green broke,
and 0.9.7 is still red.

So, this is the only branch/release currently passing, my ci tests.
There are a few minor differences in error messages, I'll have to look into.
[previously passing](https://github.com/ratmice/prop/runs/545524386?check_suite_focus=true)
[this branch](https://github.com/ratmice/prop/pull/18/checks?check_run_id=545773913)

maciejhirsz:
@ratmice great to hear. If you can either submit a PR or point me to whatever edge case regex you have that is breaking 0.10-rc2 or 0.9.7, I'm happy to have it running as a test here to ensure we don't get regressions.

ratmice:
@maciejhirsz I don't know exactly whats going on.  I had an errant `#[token = r"\p{Whitespace}"] which should be a `regex` rather than token, I haven't dug into it yet to see whats happening yet.

the 0.9.7 cases are instances fixed by #53 multi-byte reads in errors.

maciejhirsz:
There are still things I know are suboptimal in the generated code, so some more fine tuning is coming. For now though, I think things are not bad at all.

```
test identifiers                       ... bench:         701 ns/iter (+/- 30) = 1111 MB/s
test keywords_operators_and_punctators ... bench:       2,006 ns/iter (+/- 70) = 1062 MB/s
```

Going to publish this as 0.10.0-rc3, to see if the regressions people reported are gone.

PR #103, opened 2020-04-03
Callbacks and Iterators

### LINKED ISSUE #67: Provide an iterator interface
It would be nice if one could simply execute something like:
```rust
let toks: Vec<TokenData> = Token::lexer(source).iter().collect();
```
with a helper struct TokenData:
```rust
struct TokenData<'a, Token: Logos + Clone, RealSource<'a>> {
    token: Token,
        // = lexer.token.clone()
    slice: <RealSource<'a> as logos::Source<'a>>::Slice,
        // = lexer.slice()
    range: Range<usize>,
        // = lexer.range()
}
```

maciejhirsz:
This is a rough implementation of [@matklad suggestion for an ergonomic lexer](https://matklad.github.io/2018/06/06/modern-parser-generator.html#lexer), as mentioned by @CAD97 in his [reddit comment](https://www.reddit.com/r/rust/comments/fts2aw/lexer_generator_logos_010_is_finally_is_out_with/fm9ai6n/).

I've added a [test for lexing rust raw strings](https://github.com/maciejhirsz/logos/blob/ba748f522ebf506fdf5c1d481b9014ada6d45c3e/tests/tests/rust.rs#L5-L36) since that was the example brought in the blog post.

This adds a new internal trait called `Bump`, which has the following implementations:

+ `()`: no-op.
+ `usize`: straight up bump by value.
+ `Option<usize>`: bump by value, or mark token as error.

The generated code now calls `callback(lex).bump(lex)` using the new trait, which is backwards compatible with existing callbacks that return `()`.

I think it might be possible to improve UX of callbacks by abusing more generics, so that we can accept many types of callback signatures without having to resort to grabbing `&mut Lexer<Token, S>` everywhere. Two examples that immediately come to mind:

+ `fn(&mut Extras)` for whatever extras are implemented on the token.
+ `fn(match: &[u8], remainder: &[u8]) -> Bump` for this particular use-case.

---

Closes #67.


CAD97:
Ok, this is pretty awesome :)

I still wonder about the difference between the "iterator step" API (`fn lex(&str) -> (TokenKind, usize)`) versus the "stateful" API (`fn lex(&mut (&str, TokenKind)`), but I guess that doesn't matter _too_ too much.

We can probably even just provide an adapter for actually stateless lexers (i.e. ones with `Extras = ()`) via just adding another default method on `Logos`:

```rust
trait Logos {
    /* snip */
    fn first_token<'source, Source>(source: Source) -> (Self, usize)
    where
        Source: Source<'source>,
        Self: WithSource<Source>,
        Self::Extras: Meaningless,
    {
        let lexer = Self::lexer(source);
        (lexer.token, lexer.token_end)
    }
}
```

The final step for offering an alternate `librustc_lexer` would be to somehow support a data-holding enum rather than just a C-style enum. (This is, ultimately, imho, a better approach than `Extras` for most lexers, because each token kind decides what extra data it holds, and they can overlap each other. `Extras` would then be _only_ for lexer state that applies to all tokens, not just extra information that only sometimes applies to some token kinds.)

I need to go and re-check how rust-analyzer consumes rustc_lexer, and maybe pick @matklad's brain a little to try and find the ideal API shape that would help push people towards better designs while still allowing more "interesting" (read: stateful) designs to be supported.

maciejhirsz:
Pending on how far I can push generics for callbacks, you could do data holding enums either with the same attribute, or a new one like `constructor = "some_fn"`, so you could do stuff like:

```rust
fn parse_integer(slice: &str) -> i64 {
    slice.parse().unwrap()
}

enum Token {
    #[regex("[1-9][0-9]*|0", constructor = "parse_integer")]
    Integer(i64),
}
```

Could maybe do one that also takes two arguments (match and remainder) and produces the internal data + bump size, so you don't have to run over all the digits twice.

maciejhirsz:
Well, doing stuff with generics isn't super nice, I very quickly run into issues because I've to implement them over generic `F: Fn(...) -> ...` instead of concrete `fn(...) -> ...`, as any function declaration has it's own anonymous type unless deliberately cast to a function pointer. So, that's a bummer.

But, with current rust permitting arbitrary tokens in attributes, it might be possible to do inlined closures in attributes:

```rust
enum Token {
    #[regex("r\"", |lex| { ... })]
    RawString
}
```

`syn` doesn't handle arbitrary meta attributes, but it does expose the `TokenStream` of it...

CAD97:
Another option I came up with is to do more proc-macro magic:

```rust
#[logos::callback]
fn callback(/* args */) -> /* return */ { /* impl */ }
```

The transformation has to be token-based rather than type-based, but you can do fairly arbitrary transforms, or at the very least provide an argument to `logos::callback` that says what transform to do.

Then, the proc-macro expands to

```rust
/* function as written */
enum /* fn_name */ {}
impl logos::Callback for /* fn_name */ {
    fn callback(lexer: &mut Lexer) {
        /* call fn */
    }
}
```

maciejhirsz:
Ok, I just did a pretty big purge and simplified the way `Source` works, associated it with `Logos`. The downside is you won't be able to use the same `Token` enum for `&str` and `&[u8]` sources, though I'm not sure how useful that is. On the plus side, it simplifies callbacks a lot, so our raw string callback is now:

```rust
fn parse_raw_string(lexer: &mut Lexer<Token>) -> Option<usize> {
    // Who needs more then 25 hashes anyway? :)
    let q_hashes = concat!('"', "######", "######", "######", "######", "######");
    let closing = &q_hashes[..lexer.slice().len() - 1]; // skip initial 'r'

    lexer.remainder().find(closing).map(|i| i + closing.len())
}
```

I think the return value is not the right place for the bump, but it should be pretty easy to expose a safe `lexer.bump()` there. I think the return value should rather be something that can fill in the value for the enum variant.

| Return type | Produces |
|-------------|--------|
| `()` | `Token::Unit` |
| `bool` | `Token::Unit` **or** `<Token as Logos>::ERROR` |
| `Result<(), _>` | `Token::Unit` **or** `<Token as Logos>::ERROR` |
| `T` | `Token::Tuple(T)` |
| `Option<T>` | `Token::Tuple(T)` **or** `<Token as Logos>::ERROR` |
| `Result<T, _>` | `Token::Tuple(T)` **or** `<Token as Logos>::ERROR` |

So, the `parse_raw_strings` would have to look like this:

```rust
fn parse_raw_string(lexer: &mut Lexer<Token>) -> bool {
    // Who needs more then 25 hashes anyway? :)
    let q_hashes = concat!('"', "######", "######", "######", "######", "######");
    let closing = &q_hashes[..lexer.slice().len() - 1]; // skip initial 'r'

    lexer.remainder()
        .find(closing)
        .map(|i| lexer.bump(i + closing.len())
        .is_some()
}
```

And you could also do really easy callbacks like this:

```rust
#[derive(Logos)]
enum Token {
    // ...
    #[regex("-?[0-9]+", |lex| lex.slice().parse())]
    Number(i64),
}
```

maciejhirsz:
This is now a passing test :nerd_face: 

https://github.com/maciejhirsz/logos/blob/6fc92cb4094730dd2355450ba6595945d8b1cc4e/tests/tests/data.rs#L3-L35

logos/src/lexer.rs
```diff
@@ -2,14 +2,14 @@ use std::ops::Range;
...
+use crate::source::{self, Source};
 
 /// `Lexer` is the main struct of the crate that allows you to read through a
 /// `Source` and produce tokens for enums implementing the `Logos` trait.
 #[derive(Clone)]
-pub struct Lexer<Token: Logos, Source> {
+pub struct Lexer<'source, Token: Logos, Source: ?Sized> {
     /// Source from which the Lexer is reading tokens.
-    pub source: Source,
+    pub source: &'source Source,
```
ratmice: So, I'm no good at traits/deriving/etc, but
I believe the reason I no longer see an impl ToOwned, has to do with this change to Lexer.

in [ToOwned](https://doc.rust-lang.org/std/borrow/trait.ToOwned.html) there is an `impl ToOwned<T> for T where T: Clone`,
`&str` is Clone while `str` is not Clone, merely ToOwned.

So somehow this results in Lexer<'source, str> not being clone?

maciejhirsz: Ah, right. The `Clone` derive checks that all generic types are `Clone`, it's not smart enough to realize that `Source` is behind a reference which makes it not only `Clone` but `Copy`.

I've added a manual `Clone` impl for `Lexer`, so that should be fine now.

ratmice: Thanks, the testcase which caught that is [here](https://github.com/ratmice/logos/commit/f56f1762fce7cf667bf5e4a3c6ed180b608f79cc) and it works now, I had to move it lexer-mode-iterator branch, bring it in if you'd like.

maciejhirsz: Cool!

There isn't much there that's related to `logos`/`logos-derive`, but it does at least test that lexer is `Clone`, so why not :P.

maciejhirsz:
This is now documented and pretty far ahead, so gonna make a 0.11.0-rc1 release.

PR #130, opened 2020-04-23
fix: Replace type parameters nested inside types

Marwes:
https://github.com/maciejhirsz/logos/pull/128 only replaced if the
parameter were a direct field in a variant. This extends it to work if
it is inside a tuple, argument to another type etc

### DIFF
--- a/logos-derive/src/parser/type_params.rs
+++ b/logos-derive/src/parser/type_params.rs
@@ -105,7 +105,7 @@ impl TypeParams {
 }
 
 pub fn replace_lifetimes(ty: &mut Type) {
-    replace_types(ty, &mut replace_lifetime)
+    traverse_type(ty, &mut replace_lifetime)
 }
 
 pub fn replace_lifetime(ty: &mut Type) {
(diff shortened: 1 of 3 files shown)


logos-derive/src/parser/type_params.rs
```diff
@@ -142,8 +137,65 @@ pub fn replace_lifetimes(ty: &mut Type) {
 
             r.lifetime = Some(Lifetime::new("'s", span));
         },
-        Type::Slice(slice) => replace_lifetimes(&mut slice.elem),
-        Type::Tuple(tuple) => tuple.elems.iter_mut().for_each(replace_lifetimes),
         _ => (),
     }
 }
+
+pub fn replace_types(ty: &mut Type, f: &mut impl FnMut(&mut Type)) {
```
maciejhirsz, line 144: Given that this function doesn't actually replace anything by itself, would `traverse_type[s]?` (ditto for `traverse_path`) be a more fitting name?

maciejhirsz:
Published logos-derive `0.11.3` with this, `cargo update` should handle this even with logos still being `0.11.1`.

PR #131, opened 2020-04-24
Named subregex

### LINKED ISSUE #109: Feature suggestion: Named sub-regexes
As can be seen in #108 regexes can become quite complicated and long. The `regex` crate does support this feature of named character sets like `[[:digit:]]` aliasing `[0-9]`. However, not all of these are useful for everyone, and for some (or even many?) there's a lack of certain named sets. A set alone may not even be enough, but more complicated patterns may repeatedly come up here and there.

I propose a hack-y feature for `logos` that would allow users to define reüsable named sub-regexes. In my case, I have this simple rule appearing what feels like a million times:
```regex
(?x:(
  \u{000A} # LF
| \u{000B} # VT
| \u{000C} # FF
| (\u{000D} \u{000A}) # CRLF
| \u{0085} # NEL
| \u{2028} # LS
| \u{2029} # PS
))
```
Just a Unicode-aware and generous line break, with special rules like only allowing CR as combined CRLF.

The feature as I envision it would look something like this:
```Rust
#[derive(Logos)]
#[logos(named_regex("nl", r#"\u{000A} | \u{000B} | …"#))]
enum Token {
    #[error]
    Error,

    #[regex = r#"//.*{{:nl:}}"#]
    LineComment,
    …
}
```
So `#[logos(named_regex)]` defines a custom rule, given the name `nl`, followed by an arbitrary `regex`. And whenever `logos_derive` sees a `#[regex = ""]` (**Edit:** or `#[trivia]`) rule, it would scan for an occurrence of pretty much `{{:custom_rule_name:}}`, where `custom_rule_name` is the name given to `nl`. If found, replace that sub string with `(rule)` (note the braces for hygiene), where `rule` are the line breaks in my case.

This would also allow for shorter aliases than the default ones. For example, I never use `[[:xdigit:]]`, as it's just as long as `[0-9a-fA-F]`. I could alias that to just `{{:0x:}}` and do that for any number base I want.

Some suggested implementation details:

- Regex-parse the named rules before use to ensure they are complete and functional regexes, not partial ones. This penalises the happy path, though.
- I chose the `{{:x:}}` syntax for being analogous to `[[:x:]]` and for `{}` being special characters in regexes.
- During codegen, using named rules is just a matter of substring substitution.
- Round-brace-wrap the rules when doing a string replace. Technically this fucks up captures. Practically `logos` doesn't use captures.
- As implied, many sub-regexes can be defined by the user.

CAD97:
Adds the ability to write `#[logos(subpattern name = r"regex")]` on the token enum, which is then used as a "subroutine" in regex rule definitions with the syntax `(?&name)`. Closes #109.

### DIFF
--- a/logos-derive/src/parser/subpattern.rs
+++ b/logos-derive/src/parser/subpattern.rs
@@ -58,14 +58,10 @@ impl Subpatterns {
             Literal::Bytes(b) => bytes_to_regex_string(b.value()),
         };
 
-        while i < pattern.len() {
-            if let Some(f) = pattern[i..].find("(?&") {
-                i += f;
-                pattern.replace_range(i..i + 3, "(?:");
-                i += 3;
-            } else {
-                break; // done!
-            }
+        while let Some(f) = pattern[i..].find("(?&") {
+            i += f;
+            pattern.replace_range(i..i + 3, "(?:");
+            i += 3;
 
             let subref_end = if let Some(f) = pattern[i..].find(')') {
                 i + f
--- a/logos-derive/src/parser/subpattern.rs
+++ b/logos-derive/src/parser/subpattern.rs
@@ -1,52 +1,118 @@
-use syn::parse::{Parse, ParseStream};
-use syn::punctuated::{Pair, Punctuated};
-use syn::{Ident, LitStr, MetaNameValue, Token};
+use proc_macro2::TokenStream;
+use syn::{Ident, LitByteStr, LitStr};
 
-use crate::error::{Error, Result};
+use crate::error::Errors;
 use crate::mir::Mir;
+use crate::parser::definition::{bytes_to_regex_string, Literal};
 
-pub struct SubpatternInput {
-    kvs: Punctuated<MetaNameValue, Token![,]>,
+#[derive(Default)]
+pub struct Subpatterns {
+    map: Vec<(Ident, Literal)>,
 }
 
-impl Parse for SubpatternInput {
-    fn parse(input: ParseStream) -> syn::Result<Self> {
-        Ok(SubpatternInput {
-            kvs: Punctuated::parse_terminated(input)?,
-        })
-    }
-}
+impl Subpatterns {
+    pub fn add(&mut self, param: Ident, pattern: TokenStream, errors: &mut Errors) {
+        let lit = match syn::parse2::<Literal>(pattern) {
+            Ok(lit) => lit,
+            Err(e) => {
+                errors.err(e.to_string(), e.span());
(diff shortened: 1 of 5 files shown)


CAD97:
There's a "nice errors" feature that I just realized we also need: parsing/removing `(?&name)` for which we _don't_ have a definition.

This syntax is deliberately not valid for `regex-syntax`, so `regex-syntax` will error complaining about (currently) an unrecognized flag.

To avoid giving this misleading error, we should make sure to process _every_ instance of `(?&name)` syntax.

I propose:

- For a `(?&name)` that is defined but the definition has errors, replace it with `(?:name)`. (This is the same length, so a very cheap replace.)
- For a `(?&name)` that is not defined, emit an error saying the reference does not have a defined subpattern, and replace it with `(?:name)`. (This is the same length, so a very cheap replace.)

Doing the replace will allow the compilation process to proceed with a pattern that isn't perfectly accurate, but does allow compilation to continue to find any further errors without spewing misleading errors due to the earlier errors.

maciejhirsz:
> There's a "nice errors" feature that I just realized we also need: parsing/removing `(?&name)` for which we _don't_ have a definition.

Instead of iterating over all defined subpatterns for every definition, doing `while let Some(index) = regex_str.find("(?&") { ...` should be able to handle that. This will incur an extra `O(n)` cost per definition, but will avoid `O(n * m)` costs of multiple find searches for multiple subpatterns.

Bonus points if you keep track of where the last find & replace happens, that way the whole thing can be `O(n)` even if you do have to replace multiple subpatterns (ignoring subpattern lookup costs).

CAD97:
Scanning for the subpatterns to replace is now done linearly, and all subpatterns are processed.

logos-derive/src/parser/subpattern.rs
```diff
@@ -0,0 +1,118 @@
...
+        };
+
+        while i < pattern.len() {
+            if let Some(f) = pattern[i..].find("(?&") {
+                i += f;
+                pattern.replace_range(i..i + 3, "(?:");
+                i += 3;
+            } else {
+                break; // done!
+            }
```
maciejhirsz, line 68:
```suggestion
        while let Some(f) = pattern[i..].find("(?&") {
            i += f;
            pattern.replace_range(i..i + 3, "(?:");
            i += 3;
```
+ I don't think we need to worry about `i` going out of bounds here

maciejhirsz:
This is now released (logos 0.11.2, logos-derive 0.11.4, I need to get the version numbers in sync again).

PR #142, opened 2020-04-27
Add feature `regex_32` to impl Chunk trait for slices <= 32

Luro02:
This adds `regex_32`, which implements the trait for all slices from `[u8; 17]` to `[u8; 32]`. Larger variants are not possible, because std does not implement `PartialEq` and `Eq` for slices larger than 32. (`LengthAtMost32` trait)

#141


maciejhirsz:
I think it's fine to just add those without the feature flag.

Luro02:
Done

maciejhirsz:
Cheers, released as 0.11.4.

Issue #147, opened 2020-05-08
How to peek tokens and access extras?


irh:
I'm making use of `Extras` to keep track of current line information, and in my parser I also want to `peek` the next token. 

The trouble is that if I wrap the `Lexer` using `Peekable` then I no longer have a way to access the `.extras` field. 

`lexer.token` is mentioned as a solution in #82, but that doesn't seem to be available anymore. 

I assume I can make my own peekable iterator that will give me access to `.extras`, but before I go down that path I'm wondering if I'm missing something?

Thanks for the great library!

maciejhirsz:
Hey! You are not missing anything I'm afraid. I think it might be useful for `Lexer` to implement a custom `peekable` method that returns a custom peekable iterator which also derefs to `Lexer` itself so you can use all the methods (ditto for `SpannedIter`). That would probably also be a good first issue :).

irh:
OK thanks, I'll have a go at implementing it then =)

irh:
I took a quick look, and it seems like there's a potential issue with the idea. For a `Peekable` iterator we need to implement `next()` to either return the peeked value, or the next token from `Lexer::next()`. 

By auto-dereferencing `Peekable` to `Lexer` we're then making both the `Peekable::next()` and  `Lexer::next()` methods available in the same iterator (i.e. via fully qualified syntax), and mixing the use of the two with `peek()` would cause odd behaviour.

maciejhirsz:
Ah, right. Well, in that case it will just need to be boring and add a proxy method for`span`, and one for `extras` that returns `&mut T`.

xnmeet:
Is there any effective way to do this now? I have the same appeal for peek.

jeertmans:
@xnmeet did you check #151?

xnmeet:
@jeertmans I have seen this, but this example seems too simple. It ignores the situation where lexer.span will be damaged when peeking, as well as the processing related to extras that accompany it. After my attempt, wrapping the lexer does not seem to be a good way. It will make the code more complex.

xnmeet:
The problem with peek is that it really consumes the next token. At this time, the lexer cannot correctly point to information such as the span or extras associated with the current token.

jeertmans:
Hun, I see, maybe you can have a Lexer that has extras that are simply copies it the Lexer, but at previous time instants? So, even if you actually consume the lexer, you can still rollback to the previous lexer.

xnmeet:
[...]

PR #152, opened 2020-05-11
WIP - Add a peekable iterator

irh:
Following on from the discussion in #147, this PR adds a `peekable()` method to `Lexer`, which returns a custom `Peekable`-like iterator that has `peek()` as well as accessors for some of `Lexer`'s interface, namely `.extras`, `span()` and `slice()`.

### DIFF
--- a/logos/src/lexer.rs
+++ b/logos/src/lexer.rs
@@ -40,6 +40,54 @@ impl<'source, Token: Logos<'source>> Lexer<'source, Token> {
         self.source
     }
 
+    /// Wrap the `Lexer` in a peekable
+    /// [`Iterator`](https://doc.rust-lang.org/std/iter/trait.Iterator.html) that provides a
+    /// [`peek`](./struct.PeekableIter.html#method.peek) method,
+    /// which allows the next token to be inspected without it being consumed.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use logos::Logos;
+    ///
+    /// #[derive(Logos, Debug, PartialEq)]
+    /// enum Example {
+    ///     #[regex(r"[ \n\t\f]+", logos::skip)]
+    ///     #[error]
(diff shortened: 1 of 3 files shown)


irh:
@maciejhirsz Do you think this is the right direction? I haven't touched `SpannedIter`, but I could imagine adding a `PeekableSpannedIter`, and then provide `SpannedIter::peekable()` and `PeekableIter::spanned()`?

oeb25:
+1

I have some questions:
- Is there any reason not to have `peek` directly on `Lexer`?
- Should there be a way of getting span and slice on the peeked token?

irh:
>  Is there any reason not to have peek directly on Lexer?

I assume the answer to this is 'performance', you only want to pay the cost of token caching if peeking is actually needed. 

> Should there be a way of getting span and slice on the peeked token?

Well now that you mention it, if a token's been peeked, then `span` and `slice` will refer to the peeked token. I'm not sure if this is desirable behaviour or not. If not, then more caching will be needed to make the current metadata available alongside the peeked metadata.

maciejhirsz:
> * Is there any reason not to have `peek` directly on `Lexer`?

This is actually doable currently as the `Lexer` still has a field for the token internally that effectively works as a temporary. I wanted to switch the lexing to produce tokens on return value, but kept hitting performance degradation that's absent when just passing in a single `&mut` through the whole lexing processes. I'd still like to pursue that direction, since that would allow the Lexer to be polymorphic, that is to switch between different tokens depending on context. Might have to do more thinking about which direction to pursue here.

irh:
@maciejhirsz Revisiting this topic, I think it makes sense to close this PR while you consider next steps. For my purposes its sufficient to wrap the lexer with my own peeking logic, so I'm not blocked and I'm happy to continue without peeking in the lexer itself.

Issue #153, opened 2020-05-13
Non-Default extras


ExoticMatter:
It would be useful to be able to create a lexer with extras that don't implement `Default` with an alternate `Lexer` constructor that takes its `extras` as a parameter.  The `Extras: Default` constraint could then be moved to `Lexer::new` and `Logos::lexer`.

PR #154, opened 2020-05-13
Allow non-Default extras

ExoticMatter:
Solves #153.

This PR will need a semver bump, since these are breaking changes.


maciejhirsz:
Cheers!

PR #169, opened 2020-07-24
Implement Debug for Lexer

arzg:
I recently wanted to implement `Debug` for a struct that contains a `logos::Lexer`, but couldn’t. This would solve the problem.

### DIFF
--- a/logos/src/lexer.rs
+++ b/logos/src/lexer.rs
@@ -9,6 +9,7 @@ pub type Span = core::ops::Range<usize>;
 
 /// `Lexer` is the main struct of the crate that allows you to read through a
 /// `Source` and produce tokens for enums implementing the `Logos` trait.
+#[derive(Debug)]
 pub struct Lexer<'source, Token: Logos<'source>> {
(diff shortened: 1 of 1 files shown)


maciejhirsz:
This might need a manual impl since to hide some of the internals that can change in the future.

arzg:
I don’t think that’s an issue – `Debug` impls aren’t IMO part of the public interface, and it’s accepted that they can change without warning. What do you think?

Veetaha:
@maciejhirsz it's explicitly stated in the [`Stability` paragraph of `std::fmt::Debug` documentation](https://doc.rust-lang.org/stable/std/fmt/trait.Debug.html#stability) that its output is should not be considered for any stability guarantees, which is obvious because it's only used by developers to figure out what's happening under the hood without making any assumptions about what's output by `std::fmt::Debug`.

maciejhirsz:
@Veetaha thank you for the note. I still prefer to have a manual impl not to leak the internals that are not part of any of the API the user is facing. Closing in favor of #217.

PR #170, opened 2020-07-24
Fix a number of Clippy warnings

arzg:
I couldn’t fix all of them, but I did fix thirteen of them. I’d recommend adding Clippy to CI to catch these early on.


maciejhirsz (approved):

maciejhirsz:
Thanks!

PR #183, opened 2020-12-02
Enforce usage of rustfmt in CI

arzg:


arzg:
I’m going to intentionally break the formatting to see if it works, because running `cargo fmt -- --check` shows that the tests should be red.

arzg:
Ah, turns out I was using rustfmt `1.4.20-stable` locally; after running `rustup update` I am now on the latest stable, `1.4.24-stable`. This removed the diff I was seeing before that seemingly wasn’t being flagged by GitHub Actions. I’ll get rid of the purposeful style violation, since it looks like everything is working as it should.

PR #184, opened 2020-12-02
Use actions-rs/cargo instead of running cargo test manually

arzg:
This shows errors from tests using the GitHub UI, rather than just giving you the text of the error.


maciejhirsz:
Cheers!

PR #193, opened 2021-01-05
Remove unused lookup! support

### LINKED ISSUE #192: Confusing error with nonliteral explicit discriminant
https://github.com/maciejhirsz/logos/blob/ea0a3a7c2798144887243f1c535f332682091255/logos-derive/src/lib.rs#L67-L75

If a user sets a discriminant to a nonliteral expression such as `u16::MAX`, this fails with "attempt to add with overflow," which is far from indicative of the actual issue.

Instead of `.unwrap_or(usize::max_value())`, we should match the attempted evaluation, and in the case of an expression we cannot evaluate, emit a descriptive error that we don't support nonliteral explicit discriminants.

---

Side note: I think this handling is incorrect. IIUC, `Token::SIZE` should always be one larger than the largest discriminant. This is, however, demonstrably false:

```rust
enum Token {
    #[error]
    Error = 100,
    A, // = 101
    B, // = 102
}

fn main() {
    dbg!(Token::SIZE);
    dbg!(Token::B as usize);
}
```

gives `Token::SIZE = 101` and `Token::B as usize = 102`.

The simple fix here is tracking the "implicit discriminant" and checking that, rather than only the explicit discriminant.

However, it seems that `lookup!` is now completely unused by logos's internals. As such, I think the better recommendation is to remove `lookup!` and `Logos::SIZE`, and suggest users of `lookup!` use a crate such as [enum-map](https://lib.rs/crates/enum-map) instead.

CAD97:
This is entirely unused by logos's internals now, and is actually currently not correct in the face of explicit discriminants (as the discriminant of an enum variant can exceed `<Token as Logos>::SIZE`. Additionally, since we have to parse the explicit discriminant in a derive macro, said explicit discriminants are restricted to being literals, not expressions such as `u16::MAX`.

As support for `lookup!` tables is no longer necessary and is buggy and a cause of very inscrutable compiler errors (#192), I suggest that we just remove it, and suggest externals users use a different crate such as [`enum-map`](https://lib.rs/crates/enum-map) for equivalent functionality.

Fixes #192 (by removing the impacted code).

### DIFF
--- a/logos-derive/src/lib.rs
+++ b/logos-derive/src/lib.rs
@@ -24,7 +24,7 @@ use util::MaybeVoid;
 
 use proc_macro::TokenStream;
 use proc_macro2::Span;
-use quote::{quote, ToTokens};
+use quote::quote;
 use syn::spanned::Spanned;
 use syn::{Fields, ItemEnum};
 
@@ -32,7 +32,6 @@ use syn::{Fields, ItemEnum};
 pub fn logos(input: TokenStream) -> TokenStream {
     let mut item: ItemEnum = syn::parse(input).expect("Logos can be only be derived for enums");
 
-    let mut size = item.variants.len();
     let name = &item.ident;
 
     let mut error = None;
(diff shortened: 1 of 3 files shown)


maciejhirsz:
Yeah, I guess it's fine to kill it now.

Issue #196, opened 2021-01-24
Feature request: allowing ignoring case in tokens


gymore-io:
Being able to do something like that would be neat.

```rust
#[token("hello", ignore_case = true)
Hello,
```

Does an equivalent to this already exist? Would it be a good first issue? (I would like to try and implement it :) )

maciejhirsz:
I'd go for just `#[token("hello", ignore_case)]`.

It's all contained in the [proc macro here](https://github.com/maciejhirsz/logos/blob/aaa65727c71adc03e9f62e4f9c789160a4fbcb42/logos-derive/src/parser/mod.rs#L143), should be straight forward to mutate the definition given the flag, one thing to keep in mind is that this will have to work with both `token` and `regex` attributes.

gymore-io:
I started to think about the implementation but I'm running in a little issue here:

How do we discriminate between `ignore_case`, a lone ident and `ignore_case` a callback function in first position? Do I filter certain keywords `parse_callback` ?

A way to fix this issue without needing to think about callbacks may be to change the syntax into something like

```rust
#[token, ignore(case)]
Hello,
```

and just treat `ignore(case)` into a `Nested::Named("ignore", NestedValue::Group("case"))`.

Also, unrelated to the issue: would a `ignore_ascii_case` variant be useful ? Only checking ascii case is much faster may preferred in most case.

The previous syntax could also be used with this variant.

```rust
#[token, ignore(ascii_case)]
```

But I will be honest, I still prefer the "flag" version, with just `#["hello", ignore_case]`.

Any thought about that?

maciejhirsz:
Ah, you are right, a bare name assumes it's a function ref. `ignore(case)` and optionally `ignore(ascii_case)` sound good to me.

gymore-io:
#198 merged

PR #198, opened 2021-01-29
Add support for case insensitivity

gymore-io:
Issue: #196 

Case insensitivity is opt-in using the `ignore(...)` attribute.

```rust
#[derive(Logos)]
struct Token {
    /* ... */
    #[token("été", ignore(case))]
    Ete,
    #[token("DRôLE", ignore(ascii_case))]
    Drole,
    #[regex("[abc]+", ignore(case))]
    Abc,
}

let mut lex = Token::lex("été ÉTÉ DrÔle Drôle AaabbCC");
assert_eq!((Some(Token::Ete), "été"), (lex.next(), lex.slice()));
assert_eq!((Some(Token::Ete), "ÉTÉ"), (lex.next(), lex.slice()));
// Then a bunch of errors because Ô is not an ascii-uppercase ô.
assert_eq!((Some(Token::Drole), "Drôle"),  (lex.next(), lex.slice()));
assert_eq!((Some(Token::Abc), "AaabbCC"),  (lex.next(), lex.slice()));
assert_eq!(None, lex.next());
```

For now, I prevented the user to use both `ascii_case` and `case` at the same time. This can be easily changed.

--------------------------------------------

Internally, all this attribute does is converting the original literal (for tokens) or the MIR (for regular expression) into another MIR that accepts both uppercase and lowercase variants of the input.

```rust
// For example, this ...
#[token("hello", ignore(ascii_case))
Hello,

// ... is treated internally exactly like this
#[regex("[Hh][Ee][Ll][Oo]")]
Hello,
```

# Unanswered questions

* I created the `IgnoreFlags` struct to make the `ignore` attribute more easily expandable. It uses a basic bit field. I wasn't sure whether a new dependency was ok so I implemented myself what the [`bitflags`](https://crates.io/crates/bitflags) crate do (in much much simpler). Do we want a new dependency to make the code cleaner or the little features I implemented are enough?
* I am using MIR to handle case insensitivity. Is this ok or would directly creating forks be faster? If this is not acceptable/maintenable, I think it is possible to adapt the the whole thing trait to make it work. I might need a little help for that though; I found the `generator` module quite complicated to dig into and understand.
* ~Ranges (eg. `regex("[a-z]", ...)`) are currently using [`ClassUnicode::case_fold_simple`](https://docs.rs/regex-syntax/0.6.22/regex_syntax/hir/struct.ClassUnicode.html#method.case_fold_simple)  and normal literals (eg. `regex("abc", ...)`) are using Rust's [`to_lowercase`](https://doc.rust-lang.org/std/primitive.char.html#method.to_lowercase) and [`to_uppercase`](https://doc.rust-lang.org/std/primitive.char.html#method.to_uppercase) functions. This might lead to inconsistencies if those two methods do not convert characters the same way.~ Now everything uses simple folding.
 
PS: This is my first PR ever. If I've done anything wrong, please tell me! I'm both doing this because I needed this feature and to get reviews of my code. Feel free to say it needs some (or a lot) rework :)

### DIFF
--- a/logos-derive/src/lib.rs
+++ b/logos-derive/src/lib.rs
@@ -19,7 +19,7 @@ mod util;
 use generator::Generator;
 use graph::{DisambiguationError, Fork, Graph, Rope};
 use leaf::Leaf;
-use parser::{Mode, Parser};
+use parser::{ApplieIgnoreFlags, Mode, Parser};
 use util::MaybeVoid;
 
 use proc_macro::TokenStream;
@@ -132,14 +132,29 @@ pub fn logos(input: TokenStream) -> TokenStream {
                         }
                     };
 
-                    let bytes = definition.literal.to_bytes();
-                    let then = graph.push(
-                        leaf(definition.literal.span())
-                            .priority(definition.priority.unwrap_or(bytes.len() * 2))
-                            .callback(definition.callback),
-                    );
+                    if definition.ignore_flags.is_empty() {
+                        let bytes = definition.literal.to_bytes();
+                        let then = graph.push(
+                            leaf(definition.literal.span())
+                                .priority(definition.priority.unwrap_or(bytes.len() * 2))
+                                .callback(definition.callback),
+                        );
 
-                    ropes.push(Rope::new(bytes, then));
+                        ropes.push(Rope::new(bytes, then));
+                    } else {
+                        let mir = definition
+                            .literal
+                            .applie_ignore_flags(definition.ignore_flags);
+
+                        let then = graph.push(
+                            leaf(definition.literal.span())
+                                .priority(definition.priority.unwrap_or_else(|| mir.priority()))
+                                .callback(definition.callback),
+                        );
+                        let id = graph.regex(mir, then);
+
+                        regex_ids.push(id);
+                    }
                 }
                 "regex" => {
                     let definition = match parser.parse_definition(attr) {
@@ -153,7 +168,9 @@ pub fn logos(input: TokenStream) -> TokenStream {
                         .literal
                         .to_mir(&parser.subpatterns, &mut parser.errors)
                     {
-                        Ok(mir) => mir,
+                        // Calling `applie_ignore_flags` on a `Mir` when no flags are
+                        // set just returns the mir without changing it.
+                        Ok(mir) => mir.applie_ignore_flags(definition.ignore_flags),
                         Err(err) => {
                             parser.err(err, definition.literal.span());
                             continue;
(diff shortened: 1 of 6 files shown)


gymore-io:
I just noticed that  `ParserBuilder` had this neat method [`case_insensitive`](https://docs.rs/regex-syntax/0.6.22/regex_syntax/struct.ParserBuilder.html#method.case_insensitive) which basically does what I implemented with `ApplieIgnoreFlags`. But more clearly and probably more efficiently. The problem is that it does not support ascii case folding so we still have to implement that part ourselves. Plus I'm not sure how to integrate it with the current `Mir` system. Would add a parameter to `Mir::utf8` be that much of a good idea? Or maybe another constructor like `Mir::utf8_ignore_case`. I'm not exactly sure.

Moreover, using that method basically makes my `ApplieIgnoreFlags` very complicated just to turn regular expressions into ascii-case-insensitive ones and turn token literals into ascii-case-insensitive regular expressions.

I will rewrite a good portion of that part of the system, using `ParserBuilder::case_insensitive` this time.

gymore-io:
Those three last commits removed the `ApplieIgnoreCase` trait to make use of the `ParserBuilder::case_insensitive` function. This makes the code less modular (for future ignore flags for example) but more reliable to my opinion. As that function does not support ascii-case-folding, I still had to implement this part myself with a new `MakeAsciiInsensitive` trait which serves the same purpose.

logos-derive/src/lib.rs
```diff
@@ -132,14 +132,35 @@ pub fn logos(input: TokenStream) -> TokenStream {
...
+                            leaf(definition.literal.span())
+                                .priority(definition.priority.unwrap_or(bytes.len() * 2))
+                                .callback(definition.callback),
+                        );
+
+                        ropes.push(Rope::new(bytes, then));
+                    } else {
+                        let mir = definition
+                            .literal
+                            .escape_regex()
```
maciejhirsz, line 147: Is this necessary?

gymore-io: Well, as I'm calling `.to_mir` right after, we need to ensure the literal will actually be used as a string and not some regex semantics.

For example, `#[token("[Some Header]", ignore(case))]` needs to be properly escaped so that it does not turn into `Mir::Alternation('S' ... 'r')` but `Mir::Concat('[', 'S', ... , 'r', ']')`.

Originally, I had written a function that directly converts any literal into a case insensitive MIR but I later found out about the `case_insensitive` function on HIR's `ParserBuilder`. For this reason, I thought it preferable to put everything into the `to_mir` function.

maciejhirsz (approved):

maciejhirsz:
Published in 0.12

Issue #222, opened 2021-09-01
Allow lexers to use custom source types


Kixiron:
I'm currently using [`ropey`](https://docs.rs/ropey) in my LSP where `logos` is the backing lexer, and I'd like to be able to use a `Rope` natively within my generated lexer. The `Source` trait looks perfect, except there's no way I can find to be able to use arbitrary implements with `logos::Lexer`

maciejhirsz:
Fixed in #225.

kmicklas:
Unless I'm misunderstanding the `Source` API, this still isn't possible for non-contiguous types like ropes.

https://github.com/maciejhirsz/logos/pull/225#issuecomment-1484630938 notes that it might be possible using Ropey's `chunks` API, but you don't get to choose the size of those. `Source::read` lets the caller request any sized contiguous `Chunk`. The docs seem to imply that `read` should return `None` only when hitting EOF. Is that not accurate? Can we return `None` until a small enough `Chunk` is requested?

jeertmans:
@kmicklas a more recent issue #324 discussed the problem of reader by chunks, which I assume Rope to be the same. Maybe it is worth moving your question there to centralize things?

I don’t know how much related it is

kmicklas:
@jeertmans I think the concern there is a bit different since rope types should generally have no problems with random access.

kmicklas:
I put together a basic crate demoing the possibility of using Ropey and Logos together: https://github.com/kmicklas/logos-ropey

However, I can't publish it on crates.io until #332 and #333 are released.

jeertmans:
Oh that's pretty nice @kmicklas! I think Ropey's Rope is so important that it might be worth to implement that directly within the `logos` crate, to avoid creating an intermediate struct, and hide this behind some feature like `ropey`.

What do you think?

kmicklas:
> Oh that's pretty nice @kmicklas! I think Ropey's Rope is so important that it might be worth to implement that directly within the `logos` crate, to avoid creating an intermediate struct, and hide this behind some feature like `ropey`.

That would be nice! I'm happy to make a PR for that if you think it's a better approach.

jeertmans:
Super! I just merged the two PRs :)

kmicklas:
Okay, I did some testing, and https://github.com/maciejhirsz/logos/issues/222#issuecomment-1676447520 is definitely a problem. (It's not actually totally trivial to trigger a chunk size greater than 1.)

[...]

PR #225, opened 2021-09-29
Custom Source implementation when using derive macro

philliard3:
Used existing macro code as a basis for implementing a source option for the derive macro. The idea is to allow users to derive their own Source using the provided trait, which was previously missing from derive.

This work was started based on the need expressed in issue #222 . Please let me know if anything seems incorrect or anything more needs to be done.


maciejhirsz:
Better late than never, eh? :)

edko99:
Would it please be possible to implement `Source` for `ropey::Rope`?
I've tried, but got stuck because RopeSlice takes a lifetime parameter, so I cannot assign it to the `Slice` associated type.

maciejhirsz:
@edko99 I think so. Since I made the trait GATs got stabilizied, so I reckon we could change:

```
pub trait Source {
    type Slice: ?Sized + PartialEq + Eq + Debug;

    ...
}
```

To:

```
pub trait Source {
    type Slice<'source>: PartialEq + Eq + Debug;
}
```

The bigger issue is that Logos expects to be able to access continuous chunks of memory (there are hidden methods that are used just by the macro for that purpose), but I think it should still be doable by getting the first non-zero sized [chunk](https://docs.rs/ropey/latest/ropey/struct.RopeSlice.html#method.chunks) of the `RopeSlice`.

PR #231, opened 2021-12-30
Provide a way to call Lexer methods from SpannedIter

simvux:
Do you think this is appropriate or would it be better to just manually re-create some of the useful methods/impls such as `Clone` `source()` `slice()`?


marcospb19:
> would it be better to just manually re-create some of the useful methods

I prefer this approach because someone might add some method to `Lexer`, and forget to add it to `SpannedIter`.

I'd love this to be merged, ping @maciejhirsz.

simvux:
I've accidentally done some other stuff on the same branch since I forgot about this PR. But if the project wakes up again then I'll just make a new PR.

maciejhirsz:
Hey! Instead of `as_lexer` and `as_lexer_mut` could we implement `std::ops::Deref` and `std::ops::DerefMut` instead? That should make it possible to access all the `Lexer` methods (sans the ones that take `Lexer` by value) and fields more ergonomically.

PR #254, opened 2022-06-28
Allow `#[token("x", skip)]` on dataful variants

CAD97:
I ended up with a `#[regex("...", logos::skip)]` on a variant `Constant(String)` while reordering, and saw a ``the trait bound `logos::Skip: CallbackResult<'_, String, lexer::Token>` is not satisfied`` error. This attempts to make this case just work.

(In this case I'm attempting to keep commands in rough source order from a ocamllex file I'm porting, thus having an ignore attached to a dataful variant instead of `#[error]` as is usual for Logos.)

### DIFF
--- a/logos/src/internal.rs
+++ b/logos/src/internal.rs
@@ -95,7 +95,7 @@ impl<'s, P, E, T: Logos<'s>> CallbackResult<'s, P, T> for Result<P, E> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


CAD97:
Well ofc that's no good; what if your variant contained `logos::Skip`? 🙃

PR #262, opened 2022-08-18
Additional crate documentation about usage fundamentals.

Lokathor:
This adds a small amount of extra documentation to the crate level docs about how to get started using the crate.

### DIFF
--- a/logos/src/lib.rs
+++ b/logos/src/lib.rs
@@ -30,7 +30,7 @@
 //! * If the token variant holds data (eg, `Punctuation(char)`) you assign the data using a callback.
 //!   Put a `,` after your token or regex string,
 //!   and then either name a callback function (eg: `logos::skip`),
-//!   or write a callback expression inline (`eg: `|lex| lex.slice().chars().next().unwrap()`).
+//!   or write a callback expression inline (eg: `|lex| lex.slice().chars().next().unwrap()`).
 //!   There's more about callbacks below.
 //!
 //! ```rust
--- a/logos/src/lib.rs
+++ b/logos/src/lib.rs
@@ -34,7 +34,7 @@
 //!   There's more about callbacks below.
 //!
 //! ```rust
-//! use logos::{Logos, Lexer};
+//! use logos::Logos;
 //!
 //! #[derive(Logos, Debug, PartialEq)]
 //! enum Token {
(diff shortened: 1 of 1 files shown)


logos/src/lib.rs
```diff
@@ -19,8 +19,22 @@
...
+//!
+//! * One of the `enum` variants must be marked as the `#[error]` variant.
+//!   This is the token that is returned when lexing fails.
+//! * Mark a variant as a `#[regex("string")]` to match via a regular expression.
+//! * Mark a variant as a `#[token("string")]` to match any literal string
+//!   without having to worry about escaping special regex characters such as `+` and `*`.
+//! * If the token variant holds data (eg, `Punctuation(char)`) you assign the data using a callback.
+//!   Put a `,` after your token or regex string,
+//!   and then either name a callback function (eg: `logos::skip`),
+//!   or write a callback expression inline (`eg: `|lex| lex.slice().chars().next().unwrap()`).
```
domenicquirl, line 33:
```suggestion
//!   or write a callback expression inline (eg: `|lex| lex.slice().chars().next().unwrap()`).
```
Apologies for the unplanned drive-by review.

logos/src/lib.rs
```diff
@@ -19,8 +19,22 @@
...
+//!   without having to worry about escaping special regex characters such as `+` and `*`.
+//! * If the token variant holds data (eg, `Punctuation(char)`) you assign the data using a callback.
+//!   Put a `,` after your token or regex string,
+//!   and then either name a callback function (eg: `logos::skip`),
+//!   or write a callback expression inline (eg: `|lex| lex.slice().chars().next().unwrap()`).
+//!   There's more about callbacks below.
+//!
 //! ```rust
-//! use logos::Logos;
+//! use logos::{Logos, Lexer};
```
maciejhirsz, line 37:
```suggestion
//! use logos::Logos;
```

logos/src/lib.rs
```diff
@@ -45,7 +59,7 @@
 //! }
 //!
 //! fn main() {
-//!     let mut lex = Token::lexer("Create ridiculously fast Lexers.");
+//!     let mut lex: Lexer<_> = Token::lexer("Create ridiculously fast Lexers.");
```
maciejhirsz, line 62:
```suggestion
//!     let mut lex = Token::lexer("Create ridiculously fast Lexers.");
```

logos/src/lib.rs
maciejhirsz, line 37:
```suggestion
//! use logos::Logos;
```

maciejhirsz (changes requested):
I get that these make it easier to understand what's going on, but it might make users think that's what they _have to do_, which would be misleading. Explaining the type in a comment would be better IMHO.

PR #274, opened 2022-12-22
Remove ManuallyDrop

mattfbacon:
Seems to be unnecessary. Option::take provides the same functionality but with safe code.

### DIFF
--- a/logos/src/lexer.rs
+++ b/logos/src/lexer.rs
@@ -3,7 +3,6 @@ use super::Logos;
 use crate::source::{self, Source};
 
 use core::fmt::{self, Debug};
-use core::mem::ManuallyDrop;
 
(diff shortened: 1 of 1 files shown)


maciejhirsz:
> Option::take provides the same functionality but with safe code.

This is not exactly true, `ManuallyDrop::take` does not replace the value with `None` (as per comment), and not having it around the value means every invocation of `set` needs to call the drop impl on the token (even if it's always `None`). I'm not opposed to having it this way, but I'd need to first check benchmarks to make sure this doesn't degrade performance.

Ideally the token should be returned by, well, return value and kept on stack instead of the field on `Lexer`, but I had issues with performance on that one too (too many registers used for recursive calls is my somewhat educated guess).

PR #283, opened 2023-02-26
Impl Deref and DerefMut for SpannedIter

### LINKED ISSUE #266: Access extras on SpannedIter
There's the Extras item on the lexer, but there's no way to access that once you converted it to a `SpannedIter` via `Spanned`.
Would it be possible to add that to the `SpannedIter` as well?

maciejhirsz:
Closes #231, closes #266.

Note to self: acknowledge @simvux for #231 in release notes.


maciejhirsz:
> @maciejhirsz Looks like you forgot the note to self ;)

Ah! I did!

PR #295, opened 2023-04-13
impl<T: Logos> CallbackResult for T

### LINKED ISSUE #294: Multiple token types from one match
Occasionally, it is necessary for a callback to be able to return multiple types of tokens.  Currently, this isn't possible directly, but it's fairly simple to add (with the caveat that you must attach such callbacks to unit variants due to how Logos works.)

In fact, I have prepared a basic draft already.

ExoticMatter:
With this patch, you can return tokens directly from callbacks--**but only** for unit variants. Closes #294.

```rust
#[logos(skip r"[ \t\n\f]+")]
enum Token {
    #[regex(r"[a-zA-Z]+", |_| Token::Text)]
    #[regex(r"-?[0-9]+", |_| Token::Integer)]
    #[regex(r"-?[0-9]+\.[0-9]+", |_| Token::Float)]
    Text,
    Integer,
    Float,
}
```

### DIFF
--- a/logos/src/internal.rs
+++ b/logos/src/internal.rs
@@ -149,3 +149,58 @@ where
         }
     }
 }
+
+impl<'s, T: Logos<'s>> CallbackResult<'s, (), T> for T {
+    #[inline]
+    fn construct<Constructor>(self, _: Constructor, lex: &mut Lexer<'s, T>)
+    where
+        Constructor: Fn(()) -> T,
+    {
+        lex.set(Ok(self))
+    }
+}
+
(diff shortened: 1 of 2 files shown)


maciejhirsz:
Out of curiosity, do you have an example where that's something you'd do? It's always helpful to see user code for insights I might not have otherwise.

The PR looks fine to me by itself.

ExoticMatter:
In my case it was for Python-like formatting. Indent, Dedent, and NewLine all share the same regex.

maciejhirsz:
Ah interesting! Makes sense, and thank you for the PR!

maciejhirsz (approved):

PR #300, opened 2023-04-18
chore(book): adding examples

jeertmans:
This PR adds two examples of small projects that use Logos. I think it relevant to have those examples both in the handbook, as well as in the docs of the logos crates, this is why I added the scrape examples feature.

You can test it locally with `RUSTDOCFLAGS="--cfg docsrs" cargo +nightly doc --all-features -Zunstable-options -Zrustdoc-scrape-examples=examples --no-deps`.

It would be nice to fix the GitHub pages deployment (see https://github.com/maciejhirsz/logos/actions/runs/4693735567/jobs/8321101583) for future features, so it's easier for me and others to check that everything is fine.

Looking forward for your opinion on this PR :-)

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -23,24 +23,10 @@ concurrency:
   group: pages
   cancel-in-progress: true
 
-jobs:
-  # Test job
-  build-book:
-    runs-on: ubuntu-latest
-    steps:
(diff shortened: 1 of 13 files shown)


maciejhirsz (approved):
Nice! :rocket:

Issue #302, opened 2023-04-24
Incorrect regex


simensgreen:
```rust
#[derive(Debug, Logos)]
enum Token {
    #[regex(r"\(.*\)")]
    Unbrace
}

let mut lex = Token::lexer("(any text)"); // matches r"\(.*\)"

let result = lex.next();
assert!(matches!(result, Some(_)));     
assert!(matches!(result, Some(Ok(_)))); // fail
```

logos = "0.13.0"

simensgreen:
I have seen many issues about erroneous regular expressions. Why not use a ready-made, well-tested library such as [regex](https://lib.rs/crates/regex)?

jeertmans:
Nice catch for this bug!

I think Logos does not directly use the `regex` crate for one important reason: performances.

If you look closely at how Logos is implemented, it uses `regex-syntax` (from which `regex` also depends on) and parses the regex into some intermediate representation. As Logos makes heavily use of optimized look-up tables, it has to "re-implement" a lot of the logic behind regex to make it fit into its model. That's at least how I understand Logos.

maciejhirsz:
This is an easy fix, just match for anything that's not a closing paren: `r"\([^)]*\)"`

This is also not a bug per se.

> I have seen many issues about erroneous regular expressions. Why not use a ready-made, well-tested library such as [regex](https://lib.rs/crates/regex)?

So these are related questions. Logos never was, and never will be (see "Phasing out `#[regex]`" in [last release notes](https://github.com/maciejhirsz/logos/releases)) a fully featured regex engine.

Using the `regex` crate would absolutely kill the performance here. That's not throwing shade on `regex`, it's one of the fastest if not the fastest regex implementation on the planet. To use it as a backend for Logos however two things would happen:

1. The whole compiling of the state machine would happen at runtime (it happens at compile time currently). There is some runtime cost to that, though it's amortized by only having to do it once per lexer, however...
2. We would go from looking at each byte of input _at most_ twice (and more often enough it manages with looking at any byte just once), to looking _at least_ N times, where N is the number of token definitions, and at most N * X times, where X is some indefinite number of backtracks any given regex pattern can do. Even if the `regex` crate was so optimized that it only used one CPU cycle per byte when it looks at it, for a complex enough Lexer you are looking at order of magnitude or more performance loss.

(@jeertmans you replied just as I was typing this :upside_down_face:)

maciejhirsz:
[...]

PR #303, opened 2023-04-26
chore(codegen): error to prevent undesired behavior

jeertmans:
Hi @maciejhirsz !

This PR follows what was discussed in #302.

I added a small check in `logos-codegen` to check for `.*` or `.+` patterns. When found, an error is raised with a helpful message that redirects to a solution.

I hope the implementation is correct (especially for non-utf8 cases and the fact that I also added a check for `.+`).

# NOTES

1. I had to move the doc tests from `tests/tests/binary.rs` to `tests/src/lib.rs` because I don't think they were ever executed;
2. I think that the pattern you mentioned in https://github.com/maciejhirsz/logos/issues/302#issuecomment-1521342541, namely `r"\([^)]*\)"`, is missing an `\` to escape the inner `)`;
3. and I added new tests that assert a compile error is raised for the given cases.

### DIFF
--- a/logos-codegen/src/mir.rs
+++ b/logos-codegen/src/mir.rs
@@ -110,9 +110,13 @@ impl TryFrom<Hir> for Mir {
                 }
 
                 let kind = repetition.kind;
+                let is_dot = *repetition.hir == Hir::dot(!repetition.hir.is_always_utf8());
                 let mir = Mir::try_from(*repetition.hir)?;
 
                 match kind {
+                    RepetitionKind::ZeroOrMore | RepetitionKind::OneOrMore if is_dot => {
+                Err(r#"#[regex]: ".+" and ".*" regexes will always match everything, which is most propably not what you want. If you are looking to match everything until a specific character, you should use a capturing group. E.g., use regex r"\([^\}]\)" to match anything in between two parentheses. Read more about that here: https://github.com/maciejhirsz/logos/issues/302#issuecomment-1521342541."#.into())
+            }
                     RepetitionKind::ZeroOrOne => Ok(Mir::Maybe(Box::new(mir))),
                     RepetitionKind::ZeroOrMore => Ok(Mir::Loop(Box::new(mir))),
                     RepetitionKind::OneOrMore => {
--- a/logos-codegen/src/mir.rs
+++ b/logos-codegen/src/mir.rs
@@ -1,11 +1,20 @@
 use std::convert::TryFrom;
 
+use lazy_static::lazy_static;
 use regex_syntax::hir::{Hir, HirKind, RepetitionKind, RepetitionRange};
 use regex_syntax::ParserBuilder;
 
 pub use regex_syntax::hir::{Class, ClassUnicode, Literal};
 
 use crate::error::{Error, Result};
+
+lazy_static! {
+    /// DOT regex that matches utf8 only.
+    static ref DOT_UTF8: Hir = Hir::dot(false);
+
+    /// DOT regex that matches any byte.
+    static ref DOT_BYTES: Hir = Hir::dot(true);
+}
 
 /// Middle Intermediate Representation of the regex, built from
 /// `regex_syntax`'s `Hir`. The goal here is to strip and canonicalize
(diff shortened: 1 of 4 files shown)


logos-codegen/src/mir.rs
```diff
@@ -110,9 +110,13 @@ impl TryFrom<Hir> for Mir {
                 }
 
                 let kind = repetition.kind;
+                let is_dot = *repetition.hir == Hir::dot(!repetition.hir.is_always_utf8());
```
maciejhirsz, line 113: `Hir::dot` allocates. Can we put this into a `thread_local` or `once_cell` since it's always there?

logos-codegen/src/mir.rs
```diff
@@ -110,9 +110,13 @@ impl TryFrom<Hir> for Mir {
...
                 let mir = Mir::try_from(*repetition.hir)?;
 
                 match kind {
+                    RepetitionKind::ZeroOrMore | RepetitionKind::OneOrMore if is_dot => {
+                Err(r#"#[regex]: ".+" and ".*" regexes will always match everything, which is most propably not what you want. If you are looking to match everything until a specific character, you should use a capturing group. E.g., use regex r"\([^\}]\)" to match anything in between two parentheses. Read more about that here: https://github.com/maciejhirsz/logos/issues/302#issuecomment-1521342541."#.into())
+            }
```
maciejhirsz, line 119: Indent is off, should probably enforce `cargo fmt` in CI

logos-codegen/src/mir.rs
```diff
@@ -110,9 +110,13 @@ impl TryFrom<Hir> for Mir {
                 }
 
                 let kind = repetition.kind;
+                let is_dot = *repetition.hir == Hir::dot(!repetition.hir.is_always_utf8());
                 let mir = Mir::try_from(*repetition.hir)?;
 
                 match kind {
+                    RepetitionKind::ZeroOrMore | RepetitionKind::OneOrMore if is_dot => {
+                Err(r#"#[regex]: ".+" and ".*" regexes will always match everything, which is most propably not what you want. If you are looking to match everything until a specific character, you should use a capturing group. E.g., use regex r"\([^\}]\)" to match anything in between two parentheses. Read more about that here: https://github.com/maciejhirsz/logos/issues/302#issuecomment-1521342541."#.into())
```
maciejhirsz, line 118: Couple nits:

* `regexes will always match everything` -> `patterns will greedily consume the entire source till the end as Logos does not allow for backtracking`
* `you should use a capturing group` -> `you should use a negative character class`
* `r"\([^\}]\)"` -> `r"\([^)]*\)"` (wrong bracket inside, missing repetition, and I don't think you need to escape everything inside character classes)

Could also use a regular (non-raw) string and break this into multiple lines as in:

```rust
Err(
    "first line \
    second line \
    third line"
    .into()
)
```

is the same as

```rust
Err("first line second line third line".into())
```

but more readable.

maciejhirsz (changes requested):
Thanks for doing this!

logos-codegen/src/mir.rs
jeertmans: Yes, but as it wasn't already part of your dependencies, nor does it actually run during a program execution (only during compilation), I didn't feel like doing so. But I can easily make this change :)

logos-codegen/src/mir.rs
jeertmans: Yep I agree, but I think string formatting is also a nightly feature see https://rust-lang.github.io/rustfmt/?version=v1.5.1&search=#format_strings.

logos-codegen/src/mir.rs
jeertmans: > * `r"\([^\}]\)"` -> `r"\([^)]*\)"` (wrong bracket inside, missing repetition, [you don't need to escape everything inside character classes](https://regex101.com/r/fM5b8P/1)). Alternatively maybe pick a more readable example (if such a thing exists in regex), like: `r"'[^']*'"`

I didn't know, thanks!

I will apply requested changes later today ;-)

logos-codegen/src/mir.rs
maciejhirsz: Well, it won't auto-format them, but it won't break them either ;)

PR #317, opened 2023-06-24
feat(lib): impl Clone for SpannedIter

moenie99:
I haven't done this before, I hope it is correct

### DIFF
--- a/logos/src/lexer.rs
+++ b/logos/src/lexer.rs
@@ -227,7 +227,7 @@ pub struct SpannedIter<'source, Token: Logos<'source>> {
     lexer: Lexer<'source, Token>,
 }
 
-// deriving Clone doesn't add the Token::Extras: Clone bound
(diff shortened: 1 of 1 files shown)


jeertmans:
Hello! Thanks for your PR!

Could you use the Derive macro instead?

Please make sure to pass rustfmt and clippy :)

moenie99:
deriving clone doesn't generate the Token::Extras: Clone bound that is necessary to get a Clone instance for Lexer

jeertmans:
How yes my bad, just noticed Lexer’s clone was implemented the same way :)

Can you then fix Rustfmt?

moenie99:
i tried to figure it out but im not sure if i did it correctly as i never did this before, sorry. i did run rustfmt though

jeertmans:
Look’s perfect! Thanks :)

PR #320, opened 2023-06-26
chore(deps): bumping regex-syntax

jeertmans:
The `regex-syntax` introduces breaking changes between 0.6 and 0.8.

This PR tries to bump the regex-syntax dependency, to make **Logos** a bit more future-proof regarding that :-)

# Major changes

Since `regex-syntax` seems to do some optimization on the HIR, some patterns see their priority be affected by this.

E.g.: `"a|b"` (priority = 2) is optimized into `"[a-b]" (priority = 2)`.

Those optimizations also result in better performances for **Logos**, so I suggest to not go against them :-) See benchmark below.

As a result, this PR proposes to change the priority of classes, so that equivalent regex patterns have the same priority. This "rule" is tested in `logos-codegen/src/mir.rs`.

This is a breaking change, but will hopefully never change in the future, as regex HIRs cannot be optimized in something else than an equivalent pattern.

> NOTE: this should be documented in the release notes, as well as in the book. Maybe after #319 is merged?


logos-codegen/src/mir.rs
```diff
@@ -10,12 +10,28 @@ use crate::error::{Error, Result};
...
+        hir::Literal(
+            $c.encode_utf8(&mut [0; 4])
+                .as_bytes()
+                .to_vec()
+                .into_boxed_slice(),
+        )
+    };
+}
+
+pub(crate) use literal;
```
hellow554, line 33: Is there a reason for the macro to be defined here, exported but only used in `ignore_flags.rs`?

logos-codegen/src/parser/ignore_flags.rs
```diff
@@ -229,17 +229,14 @@ pub mod ascii_case {
...
         }
     }
 
     impl MakeAsciiCaseInsensitive for hir::Literal {
         fn make_ascii_case_insensitive(self) -> Mir {
-            match self {
-                hir::Literal::Byte(b) => b.make_ascii_case_insensitive(),
-                hir::Literal::Unicode(c) => c.make_ascii_case_insensitive(),
-            }
+            self.0[0].make_ascii_case_insensitive()
```
hellow554, line 239: do you think a `debug_assert_eq!(self.0.len(), 0)` would be a good idea here?

logos-codegen/src/mir.rs
```diff
@@ -10,12 +10,28 @@ use crate::error::{Error, Result};
 
 lazy_static! {
     /// DOT regex that matches utf8 only.
-    static ref DOT_UTF8: Hir = Hir::dot(false);
+    static ref DOT_UTF8: Hir = Hir::dot(Dot::AnyChar);
 
     /// DOT regex that matches any byte.
-    static ref DOT_BYTES: Hir = Hir::dot(true);
+    static ref DOT_BYTES: Hir = Hir::dot(Dot::AnyByte);
```
hellow554, line 16: I think the comments can be removed here, since `true` and `false` got a meaninful enum so the comment is not necessary anymore

logos-codegen/src/mir.rs
```diff
@@ -139,28 +154,33 @@ impl TryFrom<Hir> for Mir {
...
                     }
-                    RepetitionKind::Range(range) => match range {
-                        RepetitionRange::Exactly(n) => {
+                    // Exact {n}
+                    (n, Some(m)) if m == n => {
                             let mut out = Vec::with_capacity(n as usize);
                             for _ in 0..n {
                                 out.push(mir.clone());
                             }
                             Ok(Mir::Concat(out))
```
hellow554, line 171: this and the following has a wrong indention. Not sure why rustfmt doesn't catch this.
Could you manually decremtn the ident of these three match arms?

logos-codegen/src/mir.rs
jeertmans: I though it was used in both modules, but it maybe is not the case anymore

logos-codegen/src/mir.rs
jeertmans: Indeed :)

logos-codegen/src/mir.rs
jeertmans: Nice catch!

hellow554:
Funny enough: I also tried to update all dependencies in the logo eco system (not just regex-syntax).
Now only the test fail for me and this is two folded.

First, literals are not just one byte/char at a time, but instead a whole string/byte slice.

Therefore the following line should be changed as following:
https://github.com/maciejhirsz/logos/blob/e2702f6a742a48061cd060081f03d5297286cd2b/logos-codegen/src/mir.rs#L75

`Mir::Literal(lit) => 2 * lit.0.len()`

Second, regex-syntax does some crazy optimization, e.g. `a|b` is not an alternation, but instead a class, e.g. `a..=b` and this is correct. Even `a|c` is not an alternation, but a Class of `a..=a, c..=c`, which is all correct, but this means that some calculations are in the test cases are "wrong".

I'm not entirely sure, if they should be dropped or relaxed, but this works:

```diff
--- a/logos-codegen/src/graph/regex.rs
+++ b/logos-codegen/src/graph/regex.rs
@@ -207,7 +206,7 @@ mod tests {

         let mir = Mir::utf8("a|b").unwrap();

-        assert_eq!(mir.priority(), 2);
+        assert!((1..=2).contains(&mir.priority()));

diff --git a/logos-codegen/src/mir.rs b/logos-codegen/src/mir.rs
index 18a6307..5f258e8 100644
--- a/logos-codegen/src/mir.rs
+++ b/logos-codegen/src/mir.rs
@@ -207,17 +204,17 @@ mod tests {
     #[test]
     fn priorities() {
         let regexes = [
-            ("[a-z]+", 1),
-            ("a|b", 2),
-            ("a|[b-z]", 1),
-            ("(foo)+", 6),
-            ("foobar", 12),
-            ("(fooz|bar)+qux", 12),
+            ("[a-z]+", (1..=1)),
+            ("a|b", (1..=2)),
+            ("a|[b-z]", (1..=1)),
+            ("(foo)+", (6..=6)),
+            ("foobar", (12..=12)),
+            ("(fooz|bar)+qux", (12..=12)),
         ];

         for (regex, expected) in regexes.iter() {
             let mir = Mir::utf8(regex).unwrap();
-            assert_eq!(mir.priority(), *expected);
+            assert!(expected.contains(&mir.priority()));
         }
     }
 }

File: logos-codegen/src/mir.rs [hunk 5]
```diff
@@ -69,7 +69,12 @@ impl Mir {
             Mir::Concat(concat) => concat.iter().map(Mir::priority).sum(),
             Mir::Alternation(alt) => alt.iter().map(Mir::priority).min().unwrap_or(0),
             Mir::Class(_) => 1,
-            Mir::Literal(_) => 2,
+            Mir::Literal(lit) => {
+                match std::str::from_utf8(&lit.0) {
+                    Ok(s) => 2 * s.len(),
+                    Err(_) => 2 * lit.0.len(),
+                }
```
hellow554: I guess you're doing this because of unicode strings?
Is it possible to mix string and byte pattern in logos?
If so, can you provide a testcase where this is covered?

If not, I think doing the simple solution, e.g. `2 * lit.0.len()` is enough here. But I don't have the last word of course ;)

jeertmans:
> Funny enough: I also tried to update all dependencies in the logos eco system (not just regex-syntax). Now only the test fail for me and this is two folded.
> 
> First, literals are not just one byte/char at a time, but instead a whole string/byte slice.
> 
> Therefore the following line should be changed as following:
> 
> https://github.com/maciejhirsz/logos/blob/e2702f6a742a48061cd060081f03d5297286cd2b/logos-codegen/src/mir.rs#L75
> 
> `Mir::Literal(lit) => 2 * lit.0.len()`
> 
> Second, regex-syntax does some crazy optimization, e.g. `a|b` is not an alternation, but instead a class, e.g. `a..=b` and this is correct. Even `a|c` is not an alternation, but a Class of `a..=a, c..=c`, which is all correct, but this means that some calculations are in the test cases are "wrong".
> 
> I'm not entirely sure, if they should be dropped or relaxed, but this works:
> 
> ```diff
> --- a/logos-codegen/src/graph/regex.rs
> +++ b/logos-codegen/src/graph/regex.rs
> @@ -207,7 +206,7 @@ mod tests {
> 
>          let mir = Mir::utf8("a|b").unwrap();
> 
> -        assert_eq!(mir.priority(), 2);
> +        assert!((1..=2).contains(&mir.priority()));
> 
> diff --git a/logos-codegen/src/mir.rs b/logos-codegen/src/mir.rs
> index 18a6307..5f258e8 100644
> --- a/logos-codegen/src/mir.rs
> +++ b/logos-codegen/src/mir.rs
> @@ -207,17 +204,17 @@ mod tests {
>      #[test]
>      fn priorities() {
>          let regexes = [
> -            ("[a-z]+", 1),
> -            ("a|b", 2),
> -            ("a|[b-z]", 1),
> -            ("(foo)+", 6),
> -            ("foobar", 12),
> -            ("(fooz|bar)+qux", 12),
> +            ("[a-z]+", (1..=1)),
> +            ("a|b", (1..=2)),
> +            ("a|[b-z]", (1..=1)),
> +            ("(foo)+", (6..=6)),
> +            ("foobar", (12..=12)),
> +            ("(fooz|bar)+qux", (12..=12)),
>          ];
> 
>          for (regex, expected) in regexes.iter() {
>              let mir = Mir::utf8(regex).unwrap();
> -            assert_eq!(mir.priority(), *expected);
> +            assert!(expected.contains(&mir.priority()));
>          }
>      }
>  }
> ```

Oh yeah... this optimization is quite annoying, and seems to be caused by the AST to HIR translator, see https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6e993bda9d1d375ea2dd0bd3927a38a6.

If think there are three possible solutions to this:

+ revert back to regex-syntax v0.6 and never upgrade;
+ update the class' priority to 2, which is a breaking change, but satisfy the rule that "equivalent regexes the same priority". I worked a bit on that, but a lot of tests must be updated;
+ use regex-syntax's AST, which seems to be not optimized, instead of the HIR, or use a custom AST to HIR translator (I am not sure that optimization can be turned off).

It is also important to check whether regex-syntax's optimizations are actually impacting (or not) Logos' performances :'-)

jeertmans:
Ok so I did [a small benchmark](https://github.com/jeertmans/logos/blob/5bc523daaec7d90270d2ac00902654511027bcdb/tests/benches/bench_optimizations.rs) to compare equivalent patterns, and it seems that optimizing alternations into classes (at least a range) can improve performances, see this example:

image

hellow554:
> can improve performances

Yeah I thought so ;) Why should they do that, if not for performance :D

I mean, honestly: There's no reason for us so stick to "old" behavoir. We just need to make sure, that priorities do not conflict with each other. And I think we can do this with more tests, but IMHO we shouldn't block the upgrade just because old tests are not working.

jeertmans:
> > can improve performances
> 
> Yeah I thought so ;) Why should they do that, if not for performance :D

True! But optimizations in the `regex-syntax` crate do not necessarily translate into performance gains with `Logos`, so testing is always good ;-)

> 
> I mean, honestly: There's no reason for us so stick to "old" behavoir. We just need to make sure, that priorities do not conflict with each other. And I think we can do this with more tests, but IMHO we shouldn't block the upgrade just because old tests are not working.

I agree, but this is a major change, and hopefully we should be able to produce a stable result such that "equivalent regex patterns have the same priority". Maybe it is as simple as multiplicating the class priority by 2 ^^'

Anyway, I'd like to have @maciejhirsz's opinion on that too.

logos-codegen/src/mir.rs
jeertmans: Yes, I should come with a test for that :-)

But I did that because `Mir::Literal` can be constructed either from a `char` or from an `u8`. But a single `char`, which should have priority `2`, can be made of up to 4 bytes. So using `2 * lit.0.len()` for such a `char` could make the priority much higher. I will try to write some tests for that :-)

jeertmans: Ohh I made the dummy mistake, but `str`'s length should be `s.chars().count()` :'-)

hellow554: But then we talk about grapheme clusters? I mean `ẽ` could either be just `ẽ` or `e~`.
Please take a look at: https://stackoverflow.com/questions/58770462/how-to-iterate-over-unicode-grapheme-clusters-in-rust

hellow554:
Your commits seem to be messed up. You have duplicate commits.

Could you rebase your commits, or maybe squash them (I don't like squashing, but it's okay to use).

logos-codegen/src/mir.rs
```diff
@@ -71,8 +68,11 @@ impl Mir {
             Mir::Empty | Mir::Loop(_) | Mir::Maybe(_) => 0,
             Mir::Concat(concat) => concat.iter().map(Mir::priority).sum(),
             Mir::Alternation(alt) => alt.iter().map(Mir::priority).min().unwrap_or(0),
-            Mir::Class(_) => 1,
-            Mir::Literal(_) => 2,
+            Mir::Class(_) => 2,
```
hellow554, line 71: Maybe we need to adjust `Class` to something different.
There is the `iter` function on [`ClassUnicode`](https://docs.rs/regex-syntax/latest/regex_syntax/hir/struct.ClassUnicode.html#method.iter) and those items have a [`len` method](https://docs.rs/regex-syntax/latest/regex_syntax/hir/struct.ClassUnicodeRange.html#method.len).
Maybe that is what we want instead of a hard coded class?! I'm not really sure tbh.

tests/tests/ignore_case.rs
hellow554, line 1: Yeah, I'm not really sure what you need to adjust all the priorities.
Can you explain that please?

logos-codegen/src/mir.rs
jeertmans: We talk about unicode chars, see [regex#54](https://github.com/rust-lang/regex/issues/54).

I don't think `regex-syntax` or `logos` supports grapheme clusters, so I don't see why we should support them now?

logos/src/lib.rs
```diff
@@ -231,7 +231,7 @@ pub trait Logos<'source>: Sized {
 /// enum Token<'a> {
 ///     // We will treat "abc" as if it was whitespace.
 ///     // This is identical to using `logos::skip`.
-///     #[regex(" |abc", |_| Skip)]
+///     #[regex(" |abc", |_| Skip, priority = 3)]
```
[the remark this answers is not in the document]
jeertmans: Nope, this is because the next pattern has seen its priority go from 1 to 2. So we increase this priority to have the same behaviour as before.

logos-codegen/src/mir.rs
jeertmans: I don't think it will have a different result: a non-empty class means that at it must match one element from the class. All "special" characters are escaped in a class, and you can only match one byte or one char → priority is 2

tests/tests/ignore_case.rs
jeertmans: Because bumping `Mir::Class`'s priority from 1 to 2 broke a lot of tests or regexes, and we now need to specify the priority in such cases

PR #321, opened 2023-06-26
fix(ci): allow commenting on PR from fork

jeertmans:


jeertmans:
Ok the error seems impossible to fix without using an authorized GitHub application, that has access to the repo to write on the PR. Closing for now

PR #329, opened 2023-07-28
Update dependencies

therealbnut:
This PR updates a few of the dependencies, so that upstream fixes can be imported.

### DIFF
--- a/logos-codegen/src/mir.rs
+++ b/logos-codegen/src/mir.rs
@@ -1,7 +1,7 @@
 use std::convert::TryFrom;
 
 use lazy_static::lazy_static;
-use regex_syntax::hir::{Hir, HirKind, RepetitionKind, RepetitionRange};
+use regex_syntax::hir::{Dot, Hir, HirKind, Repetition};
 use regex_syntax::ParserBuilder;
 
 pub use regex_syntax::hir::{Class, ClassUnicode, Literal};
@@ -10,10 +10,10 @@ use crate::error::{Error, Result};
 
 lazy_static! {
     /// DOT regex that matches utf8 only.
-    static ref DOT_UTF8: Hir = Hir::dot(false);
+    static ref DOT_UTF8: Hir = Hir::dot(Dot::AnyChar);
 
     /// DOT regex that matches any byte.
-    static ref DOT_BYTES: Hir = Hir::dot(true);
+    static ref DOT_BYTES: Hir = Hir::dot(Dot::AnyByte);
 }
 
 /// Middle Intermediate Representation of the regex, built from
@@ -28,7 +28,7 @@ pub enum Mir {
     Concat(Vec<Mir>),
     Alternation(Vec<Mir>),
     Class(Class),
-    Literal(Literal),
+    Literal(char),
 }
 
 impl Mir {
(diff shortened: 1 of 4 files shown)


logos-codegen/src/mir.rs
```diff
@@ -111,23 +111,38 @@ impl TryFrom<Hir> for Mir {
 
                 Ok(Mir::Alternation(alternation))
             }
-            HirKind::Literal(literal) => Ok(Mir::Literal(literal)),
+            HirKind::Literal(literal) => {
+                let s = std::str::from_utf8(&*literal.0).unwrap();
```
therealbnut, line 115: Note, `HirKind::Literal` is now multiple characters. I'm not sure if it's required to be able to parse as utf8 or not, this may need to use a different approach if it's allowed arbitrary bytes.

logos-codegen/src/mir.rs
```diff
@@ -192,7 +194,7 @@ mod tests {
     fn priorities() {
         let regexes = [
             ("[a-z]+", 1),
-            ("a|b", 2),
+            ("a|b", 1),
```
therealbnut, line 197: Note, the HIR representation here changes from `Alternate(['a', 'b'])` to `Class('a'..='b')`, which is probably the result of an upstream optimisation - if the test wants it to be an alternate, then it's probably better to do `a|c`.

therealbnut:
Ah oops, I just noticed there's another PR that does the same - although I seem to have a smaller diff https://github.com/maciejhirsz/logos/pull/320 - it may be there's good bits from both we can take

logos-codegen/src/mir.rs
```diff
@@ -111,23 +111,38 @@ impl TryFrom<Hir> for Mir {
 
                 Ok(Mir::Alternation(alternation))
             }
-            HirKind::Literal(literal) => Ok(Mir::Literal(literal)),
+            HirKind::Literal(literal) => {
+                let s = std::str::from_utf8(&*literal.0).unwrap();
+                let mut chars = s.chars().map(Mir::Literal).peekable();
+                let c = chars.next().expect("a literal cannot be empty");
+                if chars.peek().is_some() {
+                    Ok(Mir::Concat(std::iter::once(c).chain(chars).collect()))
```
therealbnut, line 119: I suspect this is why my PR differs to #320, and doesn't require changing the class priority (producing multiple `Mir::Literal` for one `HirKind::Literal`).

logos-codegen/src/mir.rs
jeertmans: As discussed in #320, I don't think we just adapt our code to possible optimization. Otherwise, this can be a never ending game where each optimizer in `regex-syntax` would lead to a breaking change in `logos` :'-)

jeertmans: Hum i don’t think optimisations need to be lossy. To me, an optimisation results in the same behavior, which is what the priority should compute

therealbnut: I clarified what I meant as lossy as “non-invertible”. I mean that optimisations in regex-syntax may optimise in a way that we don’t have enough information to get back to the form we expect.

However we should _always_ be able to get to the same canonical DFA.

therealbnut: https://en.m.wikipedia.org/wiki/DFA_minimization

PR #332, opened 2023-08-13
Use lifetime GAT for Source::Slice

kmicklas:
This is required to write a `Source` impl for any type whose slice is a custom type and not a bare reference. For example, see [`ropey::RopeSlice`](https://docs.rs/ropey/latest/ropey/struct.RopeSlice.html).

Relates to #222.

### DIFF
--- a/logos/Cargo.toml
+++ b/logos/Cargo.toml
@@ -10,7 +10,7 @@ keywords = ["lexer", "lexical", "tokenizer", "parser", "no_std"]
 categories = ["parsing", "text-processing"]
 readme = "../README.md"
 edition = "2021"
-rust-version = "1.62.1"
+rust-version = "1.65.0"
 
 [dependencies]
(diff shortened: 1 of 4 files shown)


jeertmans:
Nice, thanks! I have read about GAT, but not actually used them.

do you have an example code that would now work with the new change, and that would not work previously?

kmicklas:
I think the change in `tests/src/lib.rs` is indicative of the types of changes that would be required in user code. In general I wouldn't expect too many incompatibilities in user code (i.e. using default `Source`s) since the current `Slice` definitions and the new ones both resolve to `&str`/`&[u8]` for the appropriate lifetime.

The bigger source of breakages would be that this invalidates any `Source` impl.

jeertmans:
Could you update Rust's MSRV so that it passes the tests? You can use `cargo-msrv` for that purpose :-)

PR #333, opened 2023-08-15
Fix ignored source attribute

kmicklas:
Currently it isn't possible to use a custom `Source` with the `Logos` derive macro. Part of #222.


jeertmans:
Thanks @kmicklas! Running the tests, and will come back to you when I have some time!

PR #334, opened 2023-08-17
Update `Source` test for GAT `Slice`

kmicklas:
#332 and #333 weren't compatible, so `master` is currently broken. Sorry I forgot to warn about that @jeertmans!


jeertmans:
Oh nice catch, that’s actually my bad I did not rebase the branch!

I see that benchmarks are failing, maybe it is worth updating the lock file?

kmicklas:
It looks like a spurious download error?

jeertmans:
Indeed, I read the error messages too fast 😅

Looks perfect!

PR #335, opened 2023-08-22
Cleanup unused Source features

kmicklas:
These are just some small cleanups I came across while working on the remaining changes for #222.

### DIFF
--- a/logos-codegen/src/generator/context.rs
+++ b/logos-codegen/src/generator/context.rs
@@ -59,21 +59,12 @@ impl Context {
         self.available.saturating_sub(self.at)
     }
 
-    pub fn read_unchecked(&mut self, len: usize) -> TokenStream {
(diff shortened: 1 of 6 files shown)


kmicklas:
@jeertmans What do you think about this? The first commit is just removing dead code but the second commit is an API simplification (which shouldn't affect most users unless they implement `Source`).

marcospb19 (approved):

maciejhirsz (approved):
This seems fine to me. Still need to rewrite all of this anyway, so I don't see any reason to keep functionality in current codebase "for the future".

jeertmans (approved):
I was first afraid of loosing some features that we might need in the future, but, as @maciejhirsz mentioned, this is not a good reason to keep an unused feature just because *we might use it later*. So I'd rather make the codebase simpler by approving this PR ;-)

Thanks for your work!

PR #340, opened 2023-10-15
Add Rc<T> and Arc<T> sources

InfiniteCoder01:


jeertmans:
Nice @InfiniteCoder01! Running the checks, can you update the code to make all tests pas?

InfiniteCoder01:
@jeertmans, sorry for not reading feature list properly. Now I've fixed it (I don't think `Rc` and `Arc` appear in nostd environment)

jeertmans:
Thanks @InfiniteCoder01! Could you please add one or two tests, and a docstring on top of each class, to show very basic examples?

InfiniteCoder01:
I don't see any tests for implementation of `Source` trait for `str` and `[u8]`. Could you give me an example?
As for docs, I didn't add any structs/types, just implemented some traits.

jeertmans:
@InfiniteCoder01 Did you check https://github.com/maciejhirsz/logos/blob/master/logos/src/source.rs?

It contains implementation of `Source` for both `str` and `[u8]`.

InfiniteCoder01:
Those are implementations, but there are no docs, nor tests for them (docs only for Source trait). But there is a module-level doc comment where I've added new sources.

jeertmans:
True, but that's not because there is no documentation at the moment that it isn't worth adding some :-)

InfiniteCoder01:
In rust, you can't add docs to `impl Trait` and it's methods. But I'll add some tests then

InfiniteCoder01:
Can't really add tests until #341

jeertmans:
oh yeah, you've got a good point!

Looking at your code, I actually wonder if it just makes sense to rather implement `Source` for `AsRef<T: impl Source>`?

ikrivosheev:
@InfiniteCoder01 hello! It will be cool to add `Box`.

marcospb19 (approved):
I implemented this for `Rc` and `Arc` using a single `impl` block:

```rust
#[cfg(feature = "std")]
impl<T> Source for T
where
    T: Deref,
    <T as Deref>::Target: Source,
{
    type Slice<'a> = <T::Target as Source>::Slice<'a>
        where T: 'a;

    fn len(&self) -> usize {
        self.deref().len()
    }

    fn read<'a, Chunk>(&'a self, offset: usize) -> Option<Chunk>
    where
        Chunk: self::Chunk<'a>,
    {
        self.deref().read(offset)
    }
    
    ...
}
```

> @InfiniteCoder01 hello! It will be cool to add Box.

`Box` also implements `Deref`, so the impl above covers it too.

marcospb19:
Box also implements `Deref`, so the impl above covers it too.

I tried using `AsRef` instead of `Deref` but couldn't.

```rust
impl<T, Inner> Source for T
where
    T: AsRef<Inner>,
    Inner: Source,
{
```
```
error[E0207]: the type parameter `Inner` is not constrained by the impl trait, self type, or predicates
```

jeertmans:
Makes sense, thank you! @InfiniteCoder01  Could you please update your branch with respect to the master and resolve conflicts? So I can merge it :)

InfiniteCoder01:
Sorry for closing the PR, I'm still bad at git stuff. Should work now, tested with `cargo test` and `cargo hack check --feature-powerset --no-dev-deps`

jeertmans:
No issue, thanks for replying so quickly @InfiniteCoder01! Running CI, it seems that you should run `cargo fmt` on your code :-)

jeertmans:
Thank you for your contribution!

Issue #265, opened 2022-10-10
Lexer produces wrong tokens when more input is provided


MalteJanz:
My logos lexer implementation somehow does not match the `TK_NOT` token when there is more input (like a whitespace) after it. Instead it matches the `TK_WORD` token in that case, which should be wrong when it has a lower priority.

Reproducible example with tests:
```rust

use logos::Logos;

#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Logos)]
#[allow(non_camel_case_types)]
pub enum SyntaxKind {
    #[regex(r"[ \t]+", priority = 1)]
    TK_WHITESPACE = 0,
    #[regex(r"[a-zA-Z][a-zA-Z0-9]*", priority = 1)]
    TK_WORD,

    #[token("not", priority = 50)]
    TK_NOT,
    #[token("not in", priority = 60)]
    TK_NOT_IN,

    #[error]
    TK_UNKNOWN,
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn single_not_works() {
        let mut lexer = SyntaxKind::lexer("not");
        assert_eq!(lexer.next(), Some(SyntaxKind::TK_NOT));
    }

    #[test]
    fn word_then_not_works() {
        let mut lexer = SyntaxKind::lexer("word not");
        assert_eq!(lexer.next(), Some(SyntaxKind::TK_WORD));
        assert_eq!(lexer.next(), Some(SyntaxKind::TK_WHITESPACE));
        assert_eq!(lexer.next(), Some(SyntaxKind::TK_NOT));
    }

    #[test]
    fn but_this_does_not_work() {
        let mut lexer = SyntaxKind::lexer("not word");
        // FAILED because
        // Left:  Some(TK_WORD)
        // Right: Some(TK_NOT)
        assert_eq!(lexer.next(), Some(SyntaxKind::TK_NOT));
        assert_eq!(lexer.next(), Some(SyntaxKind::TK_WHITESPACE));
        assert_eq!(lexer.next(), Some(SyntaxKind::TK_WORD));
    }

    #[test]
    fn this_is_fine() {
        let mut lexer = SyntaxKind::lexer("not in ");
        assert_eq!(lexer.next(), Some(SyntaxKind::TK_NOT_IN));
        assert_eq!(lexer.next(), Some(SyntaxKind::TK_WHITESPACE));
    }
}
```
[...]

PR #379, opened 2024-02-19
chore(docs): Adds graph debug documentation to book

afreeland:
I ran across [this comment](https://github.com/maciejhirsz/logos/issues/265#issuecomment-1939114278) while having some issues with `regex` support.  Thought this was beneficial information and wanted to try and include it in the book so it was a little more visible to people that may be trying to debug and understand how things are working

### DIFF
--- a/book/src/debugging.md
+++ b/book/src/debugging.md
@@ -1,6 +1,6 @@
 # Debugging
 
-Gain deeper insights into your code's behavior with this debugging section.
+Instructions on how to debug your Logos lexer.
 
 ## Visualizing Logos Graph 
 
--- a/book/src/debugging.md
+++ b/book/src/debugging.md
@@ -4,11 +4,16 @@ Instructions on how to debug your Logos lexer.
 
 ## Visualizing Logos Graph 
 
-Logos works by creating a graph that gets derived from the tokens that you defined. This graph describes how the lexer moves through different states when processing input.
+Logos works by creating a graph that gets derived from
+the tokens that you defined.
+This graph describes how the lexer moves through different
+states when processing input.
 
-Hence, it may be beneficial during debugging to be able to visualize this graph, to understand how Logos will match the various tokens. 
+Hence, it may be beneficial during debugging to be able to
+visualize this graph, to understand how Logos will match the various tokens. 
 
 If we take this example:
+
 ```rust,no_run,noplayground
 use logos::Logos;
 
@@ -36,6 +41,7 @@ fn main() {
 ```
 
 Logos actually constructs a graph that contains the logic for matching tokens:
+
 ```
 graph = {
     1: ::Fast,
@@ -64,22 +70,39 @@ graph = {
     },
 }
 ```
-This graph can help us understand how our patterns are matched, and maybe understand why we have a bug at some point.
+This graph can help us understand how our patterns are matched,
+and maybe understand why we have a bug at some point.
 
-Let's get started by trying to understand how Logos is matching the `.` character, which we've tokenized as `Token::Period`.
+Let's get started by trying to understand how Logos is matching the
+`.` character, which we've tokenized as `Token::Period`.
 
-We can begin our search by looking at number `9` for the character `.`. We can see that if Logos matches a `.` it will jump `=>` to number `2`.  We can then follow that by looking at `2` which resolves to our `::Period` token. 
+We can begin our search by looking at number `9` for the character `.`.
+We can see that if Logos matches a `.` it will jump `=>` to number `2`.
+We can then follow that by looking at `2` which resolves to our `::Period` token. 
 
-Logos will then continue to look for any matches past our `.` character.  This is required in case there is potential continuation after the `.` character. Although, in the _input_ we provided there are no any additional characters, since it is the end of our input.
+Logos will then continue to look for any matches past our `.` character.
+This is required in case there is potential continuation after the `.` character.
+Although, in the *input* we provided, there are no any additional characters,
+since it is the end of our input.
 
-We also can try to identify how the token `fast` works by looking at `9`, first, and seeing that `f` will cause Logos to jump to `7`. This will then resolve the last letters of our word _fast_ by matching `ast` which jumps to `8`. Since our provided _input_ to the lexer does not include alphabetic characters after the word "fast", but rather a whitespace, the token `::Fast` will be recognized. Then, the graph will look for further potential continuation (here, `[g-z] => 4`)
+We also can try to identify how the token `fast` works by looking at `9`,
+first, and seeing that `f` will cause Logos to jump to `7`.
+This will then resolve the last letters of our word *fast* by matching `ast`
+which jumps to `8`. Since our provided _input_ to the lexer does not include
+alphabetic characters after the word "fast", but rather a whitespace,
+the token `::Fast` will be recognized.
+Then, the graph will look for further potential continuation (here, `[g-z] => 4`)
 
-### Enabling 
+## Enabling 
 
-To enable debugging output you can define a `debug` feature in your `Cargo.toml` file, like this:
+To enable debugging output you can define a `debug` feature in your
+`Cargo.toml` file, like this:
 
 ```
 // Cargo.toml
 [dependencies]
 logos = { version = "1.2.3", features = ["debug"] }
 ```
+
+Next, you can build your project with `cargo build` and
+the output will contain a debug representation of your graph(s).
--- a/book/src/debugging.md
+++ b/book/src/debugging.md
@@ -76,10 +76,10 @@ We also can try to identify how the token `fast` works by looking at `9`, first,
 
 ### Enabling 
 
-To enable this debugging output you can use the `debug` feature.
+To enable debugging output you can define a `debug` feature in your `Cargo.toml` file, like this:
 
-In your `Cargo.toml` you can
 ```
+// Cargo.toml
 [dependencies]
 logos = { version = "1.2.3", features = ["debug"] }
 ```
--- a/book/src/debugging.md
+++ b/book/src/debugging.md
@@ -1,0 +1,85 @@
+# Debugging
+
+Gain deeper insights into your code's behavior with this debugging section.
+
+## Visualizing Logos Graph 
+
+Logos works by creating a graph that gets derived from the tokens that you defined. This graph describes how the lexer moves through different states when processing input.
+
+Hence, it may be beneficial during debugging to be able to visualize this graph, to understand how Logos will match the various tokens. 
+
+If we take this example:
+```rust,no_run,noplayground
+use logos::Logos;
+
+#[derive(Debug, Logos, PartialEq)]
+enum Token {
+    // Tokens can be literal strings, of any length.
+    #[token("fast")]
+    Fast,
+
+    #[token(".")]
+    Period,
+
+    // Or regular expressions.
+    #[regex("[a-zA-Z]+")]
+    Text,
+}
+fn main() {
+    let input = "Create ridiculously fast Lexers.";
+
+    let mut lexer = Token::lexer(input);
+    while let Some(token) = lexer.next() {
+        println!("{:?}", token);
+    }
+}
+```
+
+Logos actually constructs a graph that contains the logic for matching tokens:
+```
+graph = {
+    1: ::Fast,
+    2: ::Period,
+    3: ::Text,
+    4: {
+        [A-Z] ⇒ 4,
+        [a-z] ⇒ 4,
+        _ ⇒ 3,
+    },
+    7: [
+        ast ⇒ 8,
+        _ ⇒ 4*,
+    ],
+    8: {
+        [A-Z] ⇒ 4,
+        [a-z] ⇒ 4,
+        _ ⇒ 1,
+    },
+    9: {
+        . ⇒ 2,
+        [A-Z] ⇒ 4,
+        [a-e] ⇒ 4,
+        f ⇒ 7,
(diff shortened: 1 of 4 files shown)


book/src/debug-graph.md
```diff
@@ -0,0 +1,116 @@
...
+        . ⇒ 2,
+        [A-Z] ⇒ 4,
+        [a-e] ⇒ 4,
+        f ⇒ 7,
+        [g-z] ⇒ 4,
+    },
+}
+```
+
+Lets look at `@9` for the character `.` we see that it will jump `=>` to `2`.  We can then follow that by looking at `@2` which resolves to our `::Period` token. It will then continue to look for any matches in the case there is potential continuation after the `.` character. In the _input_ we provided there is not, since it is the end of our input.
```
jeertmans, line 62:
```suggestion
This graph can help us understand how our patterns are matched, and maybe understand why we have a bug at some point.

Let's look at the last node, `9`, for the character `.`. We see that it will jump (`=>`) to `2`.  We can then follow that by looking at `2` which resolves to our `::Period` token. It will then continue to look for any matches in the case there is potential continuation after the `.` character. In the _input_ we provided, there is not, since it is the end of our input.
```

Note that I have to read a bit more of the code to feel confident about how Graph actually works.

Also, this may be a bit counter-intuitive to take the example of `.`, where this is the last bit of our input string.

book/src/debug-graph.md
```diff
@@ -0,0 +1,116 @@
...
+
+We also can try to identify how `fast` works by looking at `@9` that `f` jumps to `7`. This will then resolve the last letters of our word _fast_ by matching `ast` which jumps to `8`. Since our provided _input_ to the lexer does not include alphanumeric character after the word "fast" but rather whitespace, the token `::Fast` will be recognized. Then the graph will look for further potential continuation (ie `[g-z] => 4`)
```
jeertmans, line 64:
```suggestion
We also can try to identify how `fast` works by looking at `9`, first, and seeing that `f` makes it jump to `7`. This will then resolve the last letters of our word _fast_ by matching `ast` which jumps to `8`. Since our provided _input_ to the lexer does not include alphanumeric character after the word "fast" but rather whitespace, the token `::Fast` will be recognized. Then the graph will look for further potential continuation (i.e., `[g-z] => 4`)
```

book/src/debug-graph.md
```diff
@@ -0,0 +1,116 @@
...
+Run `cargo build` 
+
+
+### Step 4
+Now in your code that uses Logos, point your Logos dependency in `Cargo.toml` to your local repo
+```
+[dependencies]
+logos =  { path = "../path/to/logos" }
+regex = "1.10.3"
+```
```
jeertmans, line 116: I'd be happy if there was another way than cloning the repository, because this feels overly complicated for the basic user ^^'

If not already, it might be good to add a feature that enables debugging, like with Clap's debug feature: https://github.com/clap-rs/clap/blob/0c01b5558de0a1a513a2af429981099e550d1a78/Cargo.toml#L77.

book/src/debug-graph.md
jeertmans: The thing I was to know (I guess the information is somewhere in the source code) is whether we always enter the graph via its last entry, or not.

jeertmans (changes requested):
Hey! Thank you some much for your contribution! I left a few review comments that I'd like you to address before we can merge :-)

Feel free to accept or reject my suggestions!

book/src/debug-graph.md
afreeland: Totally agree about it being counter-intuitive about the `.` example.  I used the `.` since it seemed a little more straight forward on the graph to understand.  Could probably smooth over that some if I reworked some of the wording to lead into a bit more, instead of BLAM, understand this lol

afreeland:
Yeah, sorry for the kind of rushed PR...trying to sneak this in-between family time lol

Made some improvements to language and file structure based off of your recommendations.  Hopefully, a little less jarring to someone reading it.

As for the enable debugging part, I totally agree, I am new to rust but I can try to put something together that properly uses a `debug-mode` kind of flag.  Will keep it as a separate commit or could be a follow up PR if I get a chance.  Would make life much nicer if something was an easy `feature` bit to flip.

book/src/debug-graph.md
afreeland: I looked into this some, trying to add a _feature_ that could easily be toggled.  Allowing a user to run Logos with the a `debug` feature enabled it to output the graph.  

I tried adding `debug` to the `Cargo.toml` in the root of the Logos repo
```
// Cargo.toml in repo root
[features]
debug = []
```
Then built with
`cargo build --features debug`

Then attempted to use that in another project like this
```
[dependencies]
logos =  { path = "/Users/afreeland/git/af-logos", features = ["debug"] }
```

However, that doesn't appear work because the configuration actually gets used in the `logos-codegen` section. Which is down a level and with its own `Cargo.toml` file. 

The other complication is `logos-derive` actually has `logos-codegen` as a dependency...so I tried adding the `debug` feature here, to see if it would propagate down the chain

```
// logos-derive/Cargo.toml
[dependencies]
logos-codegen = {version = "0.14.0", path = "../logos-codegen", features = ["debug"]}
```

This did output the graph when running from a different package outside of Logos but it was ALWAYS enabled at this point and not propagating.

Perhaps there is a way to pass from the top down but haven't discovered it yet.

jeertmans: I actually just added a `debug` feature, so you can use that instead.

See #382.

Once merged, you'll be able to rebase your branch and use this option (instead of cloning the git and modifying the source code) :-)

jeertmans:
> Yeah, sorry for the kind of rushed PR...trying to sneak this in-between family time lol
> 
> Made some improvements to language and file structure based off of your recommendations. Hopefully, a little less jarring to someone reading it.
> 
> As for the enable debugging part, I totally agree, I am new to rust but I can try to put something together that properly uses a `debug-mode` kind of flag. Will keep it as a separate commit or could be a follow up PR if I get a chance. Would make life much nicer if something was an easy `feature` bit to flip.

No issue, I did it myself since this was an easy fix. The issue you had was that you needed to declare a `debug` feature in sub-crates. See #382 for more details :-)

book/src/debugging.md
```diff
@@ -0,0 +1,127 @@
+# Debugging
+
+Gain deeper insights into your code's behavior with this debugging section. Here you can discover techniques and tools to explore and understand your project's internals.
+
+## Visualizing Logos Graph 
+
+Logos works by creating a graph that gets derived from the tokens that you create.  This graph visualizes how the lexer moves through different states when processing input.
```
jeertmans, line 7:
```suggestion
Logos works by creating a graph that gets derived from the tokens that you defined. This graph describes how the lexer moves through different states when processing input.
```

jeertmans: Nice description!

book/src/debugging.md
```diff
@@ -0,0 +1,127 @@
...
+
+It may be beneficial during debugging to be able to visualize how Logos has generated its graph.
```
jeertmans, line 9:
```suggestion
Hence, it may be beneficial during debugging to be able to visualize this graph,
to understand how Logos will match the various tokens. 
```

book/src/debugging.md
```diff
@@ -0,0 +1,127 @@
...
+        f ⇒ 7,
+        [g-z] ⇒ 4,
+    },
+}
+```
+This graph can help us understand how our patterns are matched, and maybe understand why we have a bug at some point.
+
+Let's get started by trying to understand how Logos is matching the `.` character, which we've tokenized as `Token::Period`.
+
+We can begin our search by looking at number `@9` for the character `.`. We can see that if Logos matches a `.` it will jump `=>` to number `2`.  We can then follow that by looking at `@2` which resolves to our `::Period` token.
```
jeertmans, line 71: I am not sure where the `@` number comes from (maybe from the comment you mentioned), but it looks weird to me to add `@` before the number, when there is none in the debug message.

jeertmans: Also, you then say number `2`, then `@2`, which may be confusing

book/src/debugging.md
```diff
@@ -0,0 +1,127 @@
...
+    },
+}
+```
+This graph can help us understand how our patterns are matched, and maybe understand why we have a bug at some point.
+
+Let's get started by trying to understand how Logos is matching the `.` character, which we've tokenized as `Token::Period`.
+
+We can begin our search by looking at number `@9` for the character `.`. We can see that if Logos matches a `.` it will jump `=>` to number `2`.  We can then follow that by looking at `@2` which resolves to our `::Period` token. 
+
+Logos will then continue to look for any matches past our `.` character.  This is required in case there is potential continuation after the `.` character. Although, in the _input_ we provided there are not any additional characters, since it is the end of our input.
```
jeertmans, line 73:
```suggestion
Logos will then continue to look for any matches past our `.` character.  This is required in case there is potential continuation after the `.` character. Although, in the _input_ we provided there are no any additional characters, since it is the end of our input.
```

book/src/debugging.md
```diff
@@ -0,0 +1,127 @@
...
+
+We also can try to identify how the token `fast` works by looking at `@9`, first, and seeing that `f` will cause Logos to jump to `7`. This will then resolve the last letters of our word _fast_ by matching `ast` which jumps to `@8`. Since our provided _input_ to the lexer does not include alphabetic characters after the word "fast" but rather whitespace, the token `::Fast` will be recognized. Then the graph will look for further potential continuation (i.e., `[g-z] => 4`)
```
jeertmans, line 75:
```suggestion
We also can try to identify how the token `fast` works by looking at `@9`, first, and seeing that `f` will cause Logos to jump to `7`. This will then resolve the last letters of our word _fast_ by matching `ast` which jumps to `@8`. Since our provided _input_ to the lexer does not include alphabetic characters after the word "fast", but rather a whitespace, the token `::Fast` will be recognized. Then, the graph will look for further potential continuation (here, `[g-z] => 4`)
```

book/src/debugging.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Enabling
```
jeertmans, line 77: This and the following section should be one level above, so:

```markdown
## Enabling

### Step 1...
```

Also, maybe name the steps (instead of saying step 1, step 2). But this may change as a result of #382.

jeertmans (changes requested):
Nice work @afreeland! I left some more comments (sorry to be picky about some ^^'), but I hope you don't mind..

Note that #382 was merged and just added a new debug feature.

book/src/debugging.md
```diff
@@ -0,0 +1,127 @@
+# Debugging
+
+Gain deeper insights into your code's behavior with this debugging section. Here you can discover techniques and tools to explore and understand your project's internals.
```
marcospb19, line 3: I'd reword this, I think this is about debugging Logos's lexer behavior and not my project internals.

book/src/debugging.md
afreeland: Yeah, I was just keeping the `@` from the comment but went ahead and stripped the `@` symbol

book/src/debugging.md
afreeland: I rebased and provided a simplified example that leverages your `debug` feature.

marcospb19: I also believe that "# Enabling" should be the first section.

book/src/debugging.md
```diff
@@ -0,0 +1,85 @@
+# Debugging
+
+Gain deeper insights into your code's behavior with this debugging section.
```
marcospb19, line 3:
```suggestion
Instructions on how to debug your Logos lexer.
```

abhillman:
Users wishing to test this functionality may do so, pending this patch making it upstream to `crates.io`, with the following:

```
logos = { git = "https://github.com/maciejhirsz/logos.git#afreeland:debug-graph" }
```

abhillman:
I haven't had a chance to dive in yet, but is there a public API for fetching, manipulating, and constructing graphs exist besides using macros? Some reasons I can think of why those APIs would be useful:

**fetching**
* if a project contains multiple lexers, being able to fetch a graph and print could be easier for users than printing all graphs, as is done in this PR
* in many cases, such an API would enable a developer to easily use the `logos` macro DSL to produce a lexer file for other tools (flex and its many derivations, for example)

**manipulating/construction**:
* although this may require additional changes, this would allow for producing lexers at runtime

jeertmans:
> I haven't had a chance to dive in yet, but does logos have a runtime-accessible public API for fetching, manipulating, and constructing graphs exist besides using macros? Some reasons I can think of why those APIs would be useful:
> 
> **fetching** [related to this PR]
> 
> * if a project contains multiple lexers, being able to fetch a graph and print could be easier for users than printing all graphs, as is done in this PR
> * in many cases, such an API would enable a developer to easily use the `logos` macro DSL to produce a lexer file for other tools (flex and its many derivations, for example)
> 
> **manipulating/construction** [out of scope for this PR, yet topically related]:
> 
> * although this may require additional changes, this would allow for producing lexers at runtime
> 
> **What this PR might look like if there is/were a fetch API**
> 
> The documentation would mention something like this:
> 
> ```rust
> let graph = Token::graph();
> println!("{:#?}", graph)
> ```
> 
> Continuing to use a feature flag to enable `::graph()` (if something like it doesn't already exist) would still also probably make sense.
> 
> Thoughts?

This is true that is might be nice to expose that as part of the public API. Would you mind creating an issue for that feature request?

jeertmans (approved):

Issue #263, opened 2022-09-09
`Clone + Copy` for `Lexer` and `SpannedIter`?


cutsoy:
I have two questions regarding `Clone` and `Copy` for `Lexer` and `SpannedIter`:

- Would it be possible to add `Copy` to `Lexer` if both the token and the extras are `Copy`?
- Similarly, would it be possible to add both `Clone + Copy` to `SpannedIter`?

I'm trying to use logos in conjunction with nom and this would help with branching. Specifically, `Clone` on `SpannedIter` would make it possible to use at all, and `Copy` would make my code look a bit cleaner.

You've probably already thought about this so apologies if the answer is obvious but I can't figure out the reason behind it myself. 😊 Or maybe I'm just the first person that would benefit from that and then I could write a PR for this.

simvux:
Hello! I currently work around the limitations of ´SpannedIter´ by allowing access to the inner lexer https://github.com/maciejhirsz/logos/pull/231, feel free to use my branch. 

Unfortunately it doesn't look like this lovely lexer will get any *new* changes anytime soon.

PR #390, opened 2024-05-02
Fix `Lexer::clone` leak and UB + tests

Jakobeha:
`Lexer::clone` shouldn't clone the inner `ManuallyDrop`, because doing so clones the inner value, which is moved out in `Lexer::next`.

This causes use-after-free if the lexer is cloned after the last-returned token is dropped, especially if the token contains an overridden implementation of `Clone` (such as `Rc`) that tries to read the dropped data.

It causes a memory leak if the token contains a heap-allocated value, because cloning makes a new allocation. This allocation is in the `ManuallyDrop` and it's guaranteed to be overridden before the call to `ManuallyDrop::take`, so it's never freed.

Another thing: https://github.com/maciejhirsz/logos/issues/263 (make `Lexer` implement `Copy`) probably should be added (referencing here because it looks like the issue has been forgotten).

### DIFF
--- a/src/lexer.rs
+++ b/src/lexer.rs
@@ -194,7 +194,7 @@ where
     fn clone(&self) -> Self {
         Lexer {
             extras: self.extras.clone(),
-            token: self.token.clone(),
+            token: ManuallyDrop::new(None),
             ..*self
         }
     }
--- a/tests/tests/clone.rs
+++ b/tests/tests/clone.rs
@@ -1,0 +1,62 @@
+use std::cell::Cell;
+
+use logos::Logos as _;
+use logos_derive::Logos;
+
+#[derive(Logos, Clone, Debug, PartialEq)]
+pub enum Token {
+    #[token("a", |_| Evil::default())]
+    Evil(Evil),
+}
+
+#[derive(Debug, Default, PartialEq)]
+pub struct Evil(Box<Cell<u8>>);
+
+impl Clone for Evil {
+    fn clone(&self) -> Self {
+        self.0.set(self.0.get() + 1);
+        Self::default()
+    }
+}
+
+#[test]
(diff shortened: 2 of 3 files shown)


jeertmans:
Hello @Jakobeha, thank you for this very comprehensive analysis!

But, if I understand well, then cloning `Lexer` becomes virtually useless?

Jakobeha:
Cloning can be used as a kind of lookahead: the cloned lexer will return the exact same tokens that would be returned by the original lexer without advancing the original (provided the lexing functions are pure). One can also `clone_from` to cloned lexer on the original to "commit" the peeked tokens.

I'm not sure if it's the most efficient way to implement a lookahead (I don't know the performance vs storing an array of spans and tokens and then making `next` into a conditional), and `source` being cloned in this method is redundant, but I'm sure in most cases it's fast enough to not matter.

The same thing (creating a "lookahead" lexer) can also be done using `Lexer::with_extras(original.remainder(), original.extras)`, or to keep the same span for lexed tokens,

```rust
let mut clone = Lexer::with_extras(original.source(), original.extras);
clone.bump(original.span().end);
```

Though it's less convenient, especially if the lexer is wrapped in another data structure, because then `Clone` on the outer structure has to be manually implemented.

---

Related, I also thought about `Copy` on `Lexer`, and personally I don't think it's a good idea, because it makes it easy to unintentionally copy the lexer instead of mutably borrowing it. `Clone` doesn't have this issue because it's explicit.

maciejhirsz (approved):

maciejhirsz:
This is a pretty good catch!

FWIW whenever the mythical rewrite happens that turns all the internal gotos from functions to `match` branches over an enum in a `loop`, the `token` field can be replaced by just a variable on stack inside `next`.

PR #404, opened 2024-07-22
refactor: don't use deprecated max_value() method

LeoDog896:
only applicable clippy lint warning


jeertmans:
Thanks @LeoDog896 :-)

Issue #398, opened 2024-06-26
Auditing use of unsafe rust


pchickey:
Hi - I am new to the `logos` ecosystem.

I don't want to cast any doubt on the correctness of logos or that anyone should choose to use it - it looks like its a remarkable project that is clearly a great choice for many users.

I am upstreaming a crate which uses logos to generate a lexer ( https://github.com/bytecodealliance/wasmtime/pull/8872 ) into wasmtime. As part of accepting `logos` to be used as a transitive dependency in the wasmtime project, I need to certify that it meets `cargo-vet`'s `safe-to-deploy` criteria: https://mozilla.github.io/cargo-vet/built-in-criteria.html#safe-to-deploy .

The lexer will be handling untrusted input, and after spending some time examining the way logos codegen works, I don't feel that I can certify that the use of `unsafe` rust is sound: while I don't have any evidence that it is unsound, the code generator, and the code it generates, is too complex for me to reasonably declare that any use of the `logos` derive macro is fully sound, as would be implied by marking it as `safe-to-deploy`.

My rough understanding, from reading the code generator and the author's blog, is that performance is a huge goal of `logos`, and it has achieved very high performance. In my use case, I don't really care about performance: the inputs to the lexer are small and infrequent, and if the lexer was one or two orders of magnitude slower, that would be fine. However, I do care about correctness, to the point of being very conservative in what dependencies I can accept.

One path forward might be to have an alternative code generator for the logos macros that uses entirely safe rust. Have the logos authors ever considered this approach? Otherwise, I will have to rewrite my lexer by hand, and in doing so I will lose composability with other logos lexers in our ecosystem.

jeertmans:
Hello @pchickey! Thanks for your comment on using unsafe Rust.

I am not the author of the code itself, so I cannot really guarantee that the use of unsafe code is done safely, nor that it is actually important at every place it is used.
But I guess that @maciejhirsz knows that better than me, and could maybe give his opinion.

I think **an important first step** would be to investigate the use of the `unsafe` keyword across all crates, and maybe list them here. I know some Rust tools have been developed to this end. Second, we could analyze each use-case separately, and see if we can simply go back to a safe alternative without loss of performances.

The solution to have a `opt-in` for safe (or unsafe) Rust could be obtained via a feature, `#[cfg(feature = safe-only)]` for example, but I wonder if that is that easy to implement, or if the compile-time might increase a lot.

This question needs further investigation, and help from anyone is more than welcome!

[...]

PR #413, opened 2024-08-19
feat(lib): add `forbid_unsafe` feature to disable unsafe code

davidkern:
See Issue: #398

As discussed in the linked issue, this PR adds an `allow_unsafe` feature and gates all unsafe code on it.
This _does not_ include that feature in the default features, so the effect of this PR would be to effectively remove unsafe code from logos unless it is opted-in.

There should be a discussion (on the issue) of the right strategy to roll this out. Removing `unsafe` code immediately might not be the best option.

There is an additional issue with the `Source` trait which this PR does not currently resolve.

## Benchmarks

Throughput of the benches for this PR and the original code (in GiB/s). Method: bench was run on release build four times, alternating between toggling `allow_unsafe` on and off. No extra effort was made to quiesce the machine other than just close other desktop apps.

It'd be very beneficial to get benchmarks on other CPUs, given the surprising result for the i7.

### Mac M3 Air

| Benchmark                                                             | Safe (this PR) | Unsafe (original)  | Delta      |
| ------------------------------------------------ | -------------  | ----------------- | -------- |
| iterate/identifiers                                                    | 1.6214            |  1.6862                 | -3.84%  |
| iterate/strings                                                         | 1.2354            |  1.2479                 | -1.00%  |
| iterate/keywords_operators_and_punctators       | 2.0598           |  2.0721                  | -0.59% |
| count_ok/identifiers                                               | 1.6191             |  1.6742                 | -3.29%  |
| count_ok/keywords_operators_and_punctators  | 1.2391             |  1.2423                | -0.25%  |
| count_ok/strings                                                    | 2.0746            |  2.0782                | -0.17%   |

Disabling unsafe code does slow down throughput by a small amount, though not as much as I anticipated.

### i7-7700HQ (Alienware 13 R3)

| Benchmark                                                             | Safe (this PR) | Unsafe (original)  | Delta       |
| ------------------------------------------------ | -------------  | ----------------- | --------- |
| iterate/identifiers                                                    | 1.0751            | 0.9426                  | +14.06% |
| iterate/strings                                                         | 0.9987           | 0.9204                  | +8.51%   |
| iterate/keywords_operators_and_punctators       | 1.3997           | 1.2977                   | +7.86%   |
| count_ok/identifiers                                               | 1.0430           | 0.9526                  | +9.49%  |
| count_ok/keywords_operators_and_punctators  | 0.9497           | 0.9088                  | +4.50%  |
| count_ok/strings                                                    | 1.3772            | 1.2962                  | +6.25%   |

I did not expect this result, at all. I suspect the compiler is able to take an optimization it otherwise couldn't, but I've not investigated why yet.

## Next steps

* Some more benchmarks
* Decision on default state of feature flag or just remove unsafe
* Documentation
* Mark `Source` unsafe if unsafe code is enabled - I think I have to duplicate the entire trait under a `cfg` because you can't conditionally compile an effect (unsafe) afaik
* `read_byte_unchecked` use is emitted by the code generator, will need to bring it back and also introduce a safe alternative


jeertmans:
Thanks for your detailed PR and comment!

This work looks good and legit to me, but I think three important things are missing:

1. `allow_unsafe` should be opt-out, so maybe to set `allow_unsafe` as a default feature, or switch it to `disallow_unsafe`. This is to limit potential changes in performances that users may not expect when bumping between two versions. People that care about safety will be able to opt out of unsafe code. Maybe in the future safe code will be the default, but I think this need further investigation.
2. Automatic tests in workflows should include a version with and a version without `allow_unsafe`.
3. What you wrote should be documented in a page of Logos' book, probably talking about safety, where unsafe code is used (as you already did the job of investigating), and how to opt-out (and potential performance costs / gains).

Thanks for your work on this :-)

jeertmans:
PS: note that the tests are currently failing, so fixing them is the priority :-)

davidkern:
Oops, I had run tests only against the logos crate itself and not the workspace. I'll work on fixing those now, looks like it is failing because I entirely removed `read_byte_unchecked` thinking it was unused.

1. I'll rename the feature to `disallow_unsafe` so that it is opt-in. I think that will be more ergonomic since adding `allow_unsafe` to defaults would mean having to use `default-features = false` and then explicitly listing the other default features just to get safe-only .
2. Thanks for enabling checks on the PR. I'll add tests against `disallow_unsafe` and also add it to the benchmark run.
3. Handbook page - will do.

I'll additionally do something about the unsafety of the Source trait so safe code can't cause UB in unsafe code.

This will take a few commits to accomplish so I've changed this over to a draft until its ready for review.

davidkern:
Will name it `forbid_unsafe` for parity with Rust's `allow`/`forbid` terminology around unsafe_code.

davidkern:
This now runs tests and benchmarks against both the default feature set and the new `forbid_unsafe` feature.

The remaining work is to split the unsafe code into new UnsafeSource and UnsafeChunk traits. That will remove the conditional compilation from the existing Source and Chunk traits, which will expose only a safe interface. Implementers of a custom source may then additionally opt-in to implementing the two new unsafe traits to enable the faster code paths (taking responsibility for any UB caused by incorrect implementation).

Splitting the traits out will remove several conditional compilation attributes and will also work better with the rustdoc generator. As I have it right now, rustdoc can't document that some trait functions depend on features. Also, changing the API based on a feature is a little weird, so this approach will avoid that too.

So some refactors and some documentation and then this will be ready for review.

jeertmans:
Thanks for your work @davidkern! Please ping me back when it's ready for a review :-)

davidkern:
Will do! I've also been busy, but will have some more time to finish this up toward the end of the week.

davidkern:
@jeertmans Sorry for the delay getting back to this.

I think this just needed some documentation, which I've now added. I did try to split up the Source trait - but this was causing more issues than it seemed to be solving, so I've instead added a safety note to that trait's documentation.

In terms of merge order, the other PR fixing the benchmark testing should go in first - that will make sure this one gets a good benchmarking run after merge, which should test out both the original unsafe code as well as the new safe-only code paths.

Please let me know if you'd like to see any changes or adjustments to the documentation!

pchickey:
@davidkern thanks so much for this work! The design you arrived at does meet my needs: by requiring the `forbid_unsafe` feature in my crate I can make sure, even if other transitive dependencies just use the default features, that there is no unsafe code in the dependency tree. (If safety was the default and unsafe was feature-driven, I would not be able to make that guarantee.)

book/src/unsafe.md
```diff
@@ -0,0 +1,32 @@
...
+
+For those applications which cannot tolerate unsafe code, the feature `forbid-unsafe`
+may be enabled. This replaces unchecked accesses in the `Logos` crate with safe,
+checked alternatives which will panic on out-of-bounds access rather than cause
+undefined behavior. Additionally, code generated by the macro will not use the
+unsafe keyword, so generated code may be used in a crates using the 
+`#![forbid(unsafe_code)]` attribute.
+
+Disabling unsafe code will generally result in a slower parser overall, though rarely
+may end up slightly faster.
```
jeertmans, line 27: Maybe mention that it depends on compiler optimizations :)

book/src/unsafe.md
```diff
@@ -0,0 +1,32 @@
...
+
+There are too many variables to consider between compiler optimizations, the specific
+grammar being parsed, and the target processor to make definitive statements about
+performance of safe-only code. The automated benchmarks of this crate shows around a
+10% slowdown in safe-only code as of the time of this writing.
```
jeertmans, line 32:
```suggestion
10% slowdown in safe-only code at the time of this writing.
```

jeertmans (changes requested):
Looks great, thanks! Let’s see what @pchickey has to say about this :)

book/src/unsafe.md
davidkern: I've reworded this section on performance to entirely remove the mention of potentially better performance.

Disabling unsafe will generally reduce throughput, so I don't want to imply that users of logos should worry about testing safe versus unsafe code to see what is faster. When the compiler does produce a surprisingly faster result, it is only for a particular micro-benchmark on a specific processor and not by a dramatic amount, so that doesn't necessarily generalize to real world usage.

book/src/unsafe.md
davidkern: Thanks, corrected this.

book/src/unsafe.md
jeertmans: Great :-) I just think users should know that, if they really care about performances, they should benchmark the two versions and measure the trade-off in speed vs safety themselves.

jeertmans:
The PR looks great @davidkern! Please reformat the code so linting and formatting checks both pass. Then, I think we can merge this :-)

davidkern:
I really should add `cargo fmt` to my git pre-checks!

That should do it...

pchickey (approved):

jeertmans (approved):

jeertmans:
Looks great @davidkern, thank you for your contribution :-)

PR #414, opened 2024-08-20
chore(ci): add rustfmt to codecov & exclude fuzz from workspace

davidkern:
While adding tests for the safe-only code PR, I noticed that the code coverage action hasn't been able to complete successfully for a while.

Originally this was caused by rustfmt not being available in the container. So this installs it before running coverage.

Additionally, the new fuzz crate was interacting poorly with the other projects in the workspace, causing link errors for both the coverage action and when running `cargo build --workspace`. This is because afl crates must be built with `cargo afl build` so they can link properly. Excluding the fuzz crate from the workspace fixes both issues and also makes sure the fuzzing exports aren't inadvertently enabled on the codegen crate outside of running a fuzz.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -12,6 +12,10 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@v4
 
+    - name: Install rustfmt
+      run: |
+        rustup component add rustfmt
+    
     - name: Generate code coverage
       run: |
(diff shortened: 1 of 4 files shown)


jeertmans:
Hi @davidkern, thanks for your PR! I currently don't have enough time to review this PR, but I'll take a look at when possible :-)

jeertmans:
Sorry for the delayed review, looks great, thanks!

Issue #434, opened 2024-11-04
Latest `nightly` on edition 2024 results in compilation failure


indietyp:
The derive code generate by logos no longer compiles for edition 2024 on nightly-2024-11-04, compilation now results in the error:

```
     |
  10 | #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Logos)]
     |                                                              ^^^^^
     |                                                              |
     |                                                              `*lex` was mutably borrowed here in the previous iteration of the loop
     |                                                              first borrow later used by call
     |
  note: this call may capture more lifetimes than intended, because Rust 2024 has adjusted the `impl Trait` lifetime capture rules
    --> libs/@local/hql/syntax-jexpr/src/lexer/token_kind.rs:10:62
     |
  10 | #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Logos)]
     |                                                              ^^^^^
     = note: this error originates in the derive macro `Logos` (in Nightly builds, run with -Z macro-backtrace for more info)
  help: add a precise capturing bound to avoid overcapturing
     |
  10 | #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Logos + use<'s>)]
     |                                                                    +++++++++
```

The code resulting in the issue:

https://github.com/hashintel/hash/blob/fb84b5182d252203c5fc58eac6a68e67eb1144c7/libs/%40local/hql/syntax-jexpr/src/lexer/token_kind.rs#L10-L45

jeertmans:
Hello @indietyp, thank you for reporting!

Nightly isn’t stable, so I not rush before calling that a bug. Could you find the RFC or nightly feature that breaks the build? So I can read the details and see how likely it is going to break our code in the future.

indietyp:
Yea sure, this has something to do with edition 2024, in which the impl capture rules are changed, see: https://github.com/rust-lang/rust/issues/117587 and https://github.com/rust-lang/rfcs/pull/3498

This has regressed between 2024-10-28 and 2024-11-04.

Specifically in our example the problem was with:

```
fn callback<'s>(_: &mut Lexer<'s>) -> impl CallbackResult<'s, bool, TokenKind<'s>> {
    false
}
```

adding a `+ use<'s>` will solve this at the return type will solve this.

In edition 2024 this captures both `'s` and the anonymous lifetime of the lexer. This is intended behaviour from edition 2024 onwards.

[...]

PR #435, opened 2024-11-12
fix(lib): use precise capturing on 1.82+

indietyp:
Fixes https://github.com/maciejhirsz/logos/issues/434 by conditionally adding `+ use<'s>` to the implementation of the callback, but _only_ if on 1.82+

Additionally it adds tests for the MSRV and 1.82 to the CI.

### DIFF
--- a/.github/workflows/rustlib.yml
+++ b/.github/workflows/rustlib.yml
@@ -1,8 +1,8 @@
 on:
   pull_request:
     paths:
-    - '**.rs'
-    - '**/Cargo.toml'
+      - "**.rs"
+      - "**/Cargo.toml"
   workflow_dispatch:
 
 name: Library testing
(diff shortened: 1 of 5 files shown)


.github/workflows/rustlib.yml
jeertmans, line 1: You just probably turn off your formatter here and only apply changes ;-)

logos-codegen/build.rs
jeertmans, line 1: I think this would be great to add some comments pointing back to your original issue, so we know where it comes from :-)

jeertmans (changes requested):
Looks great, thanks for being so quick!

Please take a look a my comments.

.github/workflows/rustlib.yml
indietyp: I yea sorry my code editor kinda just did it here, will revert the change.

jeertmans:
As indicated by the CI, I think the new build dependency requires us to bump the MSRV to 1.74

indietyp:
I don't think this is actually caused by the new build-dependency (it also has `semver` as a dependency) I think the issue exists because the toolchain was never before tested (see the `{{matrix.rust}}` addition). I think this might be a problem with rstest.

logos-codegen/build.rs
indietyp: see [fc71df4](https://github.com/maciejhirsz/logos/pull/435/commits/fc71df4372c02bc06c6ad0469fa3a510613a8dba)

.github/workflows/rustlib.yml
indietyp: see [89ee9dc](https://github.com/maciejhirsz/logos/pull/435/commits/89ee9dc2f814a4d693a935d740f9a69fcc4114f2)

indietyp:
actually correction, looking at the output of `cargo tree --workspace` the issue is `logos-cli`:

https://github.com/maciejhirsz/logos/blob/491842d0adb9689e9fa707e6fb89b97f1a59fe28/logos-cli/Cargo.toml#L11

jeertmans:
> I don't think this is actually caused by the new build-dependency (it only has `semver` as a dependency) I think the issue exists because the toolchain was never before tested (see the `{{matrix.rust}}` addition). I think this might be a problem with rstest.

I see, but I don't get why the `msrv_check` workflow didn't catch this before...



> actually correction, looking at the output of `cargo tree --workspace` the issue is `logos-cli`:
> 
> https://github.com/maciejhirsz/logos/blob/491842d0adb9689e9fa707e6fb89b97f1a59fe28/logos-cli/Cargo.toml#L11

You are right, I was too quick on my conclusions :-)

logos-codegen/build.rs
jeertmans: Great thanks!

.github/workflows/rustlib.yml
jeertmans: No issue, thanks!

indietyp:
We could either raise the MSRV to 1.74, raise the CI to only test 1.74 or downgrade predicates to something that supports 1.70

jeertmans:
> We could either raise the MSRV to 1.74, raise the CI to only test 1.74 or downgrade predicates to something that supports 1.70

I think the issue is actually the CI doesn't use `--locked`, so this is probably why the issue wasn't discovered before.

For the sake of simplicity, we can just bump to 1.74 here.

Cargo.toml
```diff
@@ -30,7 +33,7 @@ rust-version.workspace = true
 version.workspace = true
 
 [package.metadata]
-msrv = "1.70.0"  # Needed to duplicate, because cargo-msrv does not support workspace
+msrv = "1.70.0" # Needed to duplicate, because cargo-msrv does not support workspace
```
jeertmans, line 36: I think you need to also update this field

jeertmans:
Looks all good @indietyp, thanks!

jeertmans (approved):

PR #441, opened 2024-11-29
Add callbacks and priority for #[logos(skip)]

### LINKED ISSUE #440: Add callback to skip
Is it possible to add a callback to skip?

This would be useful if you are trying to count newlines, and you also want to skip them:
```rust
use logos::{Lexer, Logos, Skip};

#[derive(Debug, Clone, Copy, Default)]
struct Extras {
    line_num: usize,
}

#[derive(Logos)]
#[logos(skip r" \r")]
// Is there a way of adding a callback here?
#[logos(skip(regex = r"\n", callback = newline_callback))]
#[logos(extras = Extras)]
enum Token {
    #[regex("[a-zA-Z]", |lexer| { lexer.slice().chars().next().unwrap() })]
    Letter(char),
    #[regex("[0-9]", |lexer| { lexer.slice().parse::<u8>().unwrap() })]
    Number(u8),
}

fn newline_callback(lexer: &mut Lexer<Token>) -> Skip {
    lexer.extras.line_num += 1;
    Skip
}
```
I know you could add another variant like `Token::Ignored`, or attach the newline regex to another variant, but both those solutions feel messy, and it'd be nice to have all the skip logic in one place.

mysteriouslyseeing:
Closes #440.

Syntax is same as #[regex] and #[token]:
`#[logos(skip(literal[, [callback =] callback][, priority = priority]))]`

Also added an example with usage:
```rust
fn main() {
    let mut lexer = Token::lexer("abc 123\nab( |23\nAbc 123");
    let mut out_tokens = Vec::new();

    while let Some(token_result) = lexer.next() {
        if let Ok(token) = token_result {
            out_tokens.push(token);
        } else {
            // Oh no! There was an error!
            eprintln!("Unrecognised character on line {}", lexer.extras.line_num + 1);
        }
    }
}

use logos::{Lexer, Logos, Skip};

#[derive(Debug, Clone, Copy, Default)]
struct Extras {
    line_num: usize,
}

#[logos(skip r"[ \r]")]
#[logos(skip(r"\n", callback = newline_callback, priority = 3))]
#[logos(extras = Extras)]
enum Token {
    #[regex("[a-z]+")]
    Letters,
    #[regex("[0-9]+")]
    Numbers,
}

fn newline_callback(lexer: &mut Lexer<Token>) -> Skip {
    lexer.extras.line_num += 1;
    Skip
}
```
A lot of the logic is copied and pasted from Definition - maybe it would be better to refactor them both to use common code? Let me know.


jeertmans:
Hi @mysteriouslyseeing, it looks great, thanks for your contribution!

Before I review this in depth, could you please:

- add unit tests to check that all syntax variations are supported;
- add documentation to the book so that people will know this feature is supported (you can also link to the example file you wrote)?

Thanks!

tests/tests/callbacks.rs
```diff
@@ -216,3 +216,92 @@ mod return_result_skip {
...
+    enum Token<'src> {
+        #[regex(r"<[a-zA-Z0-9-]+>", |lex| &lex.slice()[1..lex.slice().len()-1])]
+        Tag(&'src str),
+    }
+
+    fn skip_comment<'src>(lexer: &mut Lexer<'src, Token<'src>>) {
+        let end = lexer
+            .remainder()
+            .find("-->")
+            .unwrap_or(lexer.remainder().len());
```
jeertmans, line 242: In line with other comment, you can probably fix the text by changing this to `.map(|id| id + 3).unwrap_or(lexer.remainer().len());` and `bump(end)`.

jeertmans (changes requested):
Thanks for adding tests and book docs!

Before I review other parts of the code, please check my first comments about the failing test :-)

mysteriouslyseeing:
When I run `cargo test` in the `./test` subdirectory, some doctests fail. They don't appear to have anything to do with my changes. Is that something to do with me?

tests/tests/callbacks.rs
```diff
@@ -216,3 +216,92 @@ mod return_result_skip {
...
+    fn skip_callback_function() {
+        let mut lexer = Token::lexer("<foo> <!-- comment --> <bar>");
+        assert_eq!(lexer.next(), Some(Ok(Token::Tag("foo"))));
+        assert_eq!(lexer.next(), Some(Ok(Token::Tag("bar"))));
+        assert_eq!(lexer.next(), None);
+
+        let mut lexer = Token::lexer("<foo> <!-- unterminated comment");
+        assert_eq!(lexer.next(), Some(Ok(Token::Tag("foo"))));
+        // Errors not allowed from skips
+        assert_eq!(lexer.next(), None);
```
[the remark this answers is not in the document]
mysteriouslyseeing: That was to do with the other branch, where skips could return errors. I've removed that line

tests/tests/callbacks.rs
mysteriouslyseeing: Thanks!

tests/tests/callbacks.rs
```diff
@@ -216,3 +216,92 @@ mod return_result_skip {
...
+    }
```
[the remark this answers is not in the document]
mysteriouslyseeing: Whoops, I forgot to run `cargo test` in `./test`. My bad, I'll fix that now.

jeertmans:
> be confusing to people previously familiar with the codebase. I've committed it like this, b

I think this is better to avoid breaking changes when possible (and changing the name of an enum variant is a breaking change), so I prefer reverting to previous name (and IMO it conveys the functionality better), and disable Clippy warning for this variant.

jeertmans:
Hi @mysteriouslyseeing, sorry for the delay in the review!

mysteriouslyseeing:
Yes, I am still available

logos-codegen/src/parser/mod.rs
```diff
@@ -298,6 +353,23 @@ impl Parser {
...
     }
 
+    fn parse_skip_callback(&mut self, tokens: TokenStream) -> Option<SkipCallback> {
+        let span = tokens.span();
+        Some(match self.parse_callback(tokens) {
+            Some(Callback::Inline(inline)) => SkipCallback::Inline(inline),
+            Some(Callback::Label(label)) => SkipCallback::Label(label),
+            Some(Callback::Skip(_)) => {
+                // Probably not reachable
+                return None;
```
jeertmans, line 363: Can you comment in this being unreachable? If we can guarantee this, we should probably use `unreachable!`, so we convey the right message. If an error occurs, it should be during compile time, and indicate that something went wrong (instead of silently returning `None`).

jeertmans (commented):
Small comments before my approval, let me know what you think :-)

tests/tests/callbacks.rs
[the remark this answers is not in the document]
mysteriouslyseeing: I think I actually remember looking into this at some point - you would want to add a trait alongside CallbackResult (like SkipCallbackResult) and implement it for types like:

- (): a guaranteed skip
- Skip: a guaranteed skip
- Result<Skip, Error>: either a skip or an error

And then require the skip callback to return one of those types.

logos-codegen/src/parser/mod.rs
mysteriouslyseeing: The function that it calls, self.parse_callback, can only return None or Some(Callback::Inline) or Some(Callback::Label). Should that be an `unreachable!` or would a `self.err` be more appropriate?

mysteriouslyseeing: I've added a self.err, but let me know if I should change it.

jeertmans: If that should really be unreachable, then `unreachable!` is fine. If an usage error can lead to this, then `self.err` is more appropriate (because this is the user's mistake).

tests/tests/callbacks.rs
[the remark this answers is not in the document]
jeertmans: I see, makes sense. This can be another (future) PR if you don't feel adding support for `Result<Skip, Error>`, up to you :-) I feel this is not so important, as parsing errors can still be performed with normal callbacks (i.e., attached to token variants).

tests/tests/callbacks.rs
```diff
@@ -216,3 +216,85 @@ mod return_result_skip {
...
+    use super::*;
+
+    #[derive(Debug, Clone, Copy, Default)]
+    struct Extras {
+        line_num: usize,
+    }
+
+    #[derive(Logos, Debug, PartialEq)]
+    #[logos(skip r"[ \r]")]
+    #[logos(skip(r"\n", callback = |lex| { lex.extras.line_num += 1; }, priority = 3))]
```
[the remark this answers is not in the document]
mysteriouslyseeing: I've added a test that requires it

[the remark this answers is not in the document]
mysteriouslyseeing: Thinking about it, it could also be used in the brainfuck example:
If you add
```rust
// line 42
#[derive(Debug, Logos)]
// skip all non-op characters
#[logos(skip(".|\n", priority = 0))]
enum Op { /* ... */ }
```
and change
```rust
// line 89
pub fn execute(code: &str) {
    let operations: Vec<_> = Op::lexer(code).collect::<Result<_, _>>().unwrap();
    /* ... */
}
```
that could also demonstrate usage? Up to you.

jeertmans (approved):

jeertmans:
Can you take a look at CodeCov's comments? Some lines you added / modified are not covered by tests. If possible, all lines should be covered.

mysteriouslyseeing:
Just going to convert to draft while adding test coverage

mysteriouslyseeing:
Quite a lot of what codecov was talking about was codegen error conditions - what is the recommended way to test for those?

tests/tests/callbacks.rs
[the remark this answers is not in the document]
mysteriouslyseeing: I added the trait and the implementation

mysteriouslyseeing:
The failing test was due to the `+ use<'s>` bound that happens with the rust_1_82 feature.

logos-codegen/tests/codegen.rs
jeertmans, line 1: Hi @mysteriouslyseeing, I see a lot of nice changes and improvements! Regarding the check that outputs match, you could probably simplify a bit by using snapshot testing, such as provided by [insta.rs](https://insta.rs/).

mysteriouslyseeing: I've moved the tests to use insta.

PR #442, opened 2024-11-29
Minor readme improvments

tkr-sh:
- Put the release process in it's own file
- Removed extra whitespaces in a rust code block

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,48 +25,48 @@ To achieve those, **Logos**:
 ## Example
 
 ```rust
- use logos::Logos;
+use logos::Logos;
 
- #[derive(Logos, Debug, PartialEq)]
- #[logos(skip r"[ \t\n\f]+")] // Ignore this regex pattern between tokens
- enum Token {
-     // Tokens can be literal strings, of any length.
-     #[token("fast")]
-     Fast,
+#[derive(Logos, Debug, PartialEq)]
+#[logos(skip r"[ \t\n\f]+")] // Ignore this regex pattern between tokens
+enum Token {
+    // Tokens can be literal strings, of any length.
+    #[token("fast")]
+    Fast,
 
-     #[token(".")]
-     Period,
+    #[token(".")]
+    Period,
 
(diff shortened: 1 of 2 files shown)


jeertmans:
Hi @tkr-sh, thanks for your contribution!

I like when code is formatted, so cleaning up examples is always good. Did you do that manually, or did you use a tool?

Regarding the creation of another file, I don't really see the benefit. This is probably better to put this in the book, under a *Contributing* section, for example.

tkr-sh:
I personally don't think that it's that a good idea to have that in the README, and here are some other projects that use a separate file:
- https://github.com/BurntSushi/ripgrep/blob/master/RELEASE-CHECKLIST.md
- https://github.com/bootandy/dust/blob/1372815007572ab1c2a63ee753381e4b4f60e95e/ci/how2publish.txt
- https://github.com/lambda-fairy/maud

and there are probably a lot more. But, if you don't agree, it's ok and I won't argue. Putting it in a book is also ok!

Concerning the "cleaning up", I just noticed that there were spaces at the start of the line, so I just "manually" deleted them

jeertmans:
> I personally don't think that it's that a good idea to have that in the README, and here are some other projects that use a separate file:
> 
> * https://github.com/BurntSushi/ripgrep/blob/master/RELEASE-CHECKLIST.md
> * https://github.com/bootandy/dust/blob/1372815007572ab1c2a63ee753381e4b4f60e95e/ci/how2publish.txt
> * https://github.com/lambda-fairy/maud
> 
> and there are probably a lot more. But, if you don't agree, it's ok and I won't argue. Putting it in a book is also ok!

Ok that makes sense, let's keep it that way, and we will see in the future if we write a contributing guide or not :-)


> Concerning the "cleaning up", I just noticed that there were spaces at the start of the line, so I just "manually" deleted them

Ok, I guess it was introduced by mistake in the past, but they look legit, so I trust the spacing must be consistent!

Thanks for your contribution!

PR #445, opened 2024-12-02
Added initial functionality to error callback

### LINKED ISSUE #444: Add option to change error constructor
Following up from discussion in #440:
The current implementation of a token's Error type makes use of the Default constructor. This works fine as a default, but you run into issues if you want the error to reflect the current span of the Lexer, for example, to provide users of a lexer an indication of where exactly there was an error. It's not impossible - currently, a solution is to provide an arbitrary token variant with an attribute like the following:
```rust
use logos::Logos;

#[derive(Logos)]
#[logos(error = String)]
enum Token {
    #[token("a", priority = 1)]
    A,
    #[token("b", priority = 1)]
    #[regex(".", callback = |lex| {
        Err::<(), String>(format!("Syntax error at {:?}: unrecognised character '{}'", lex.span(), lex.slice()))
    })]
    B,
}
```
Note that you have to add priority = 1 to both A and B because "." also matches "a" and "b", and you also have to specify the associated type of the result because Rust cannot infer it.

A solution is to allow users to provide a default constructor for Error, to be used instead of `Default::default()` with an attribute like `#[logos(error_callback = ...)]`, or something similar. The previous example with this would look like this:
```rust
use logos::Logos;

#[derive(Logos)]
#[logos(error = String)]
#[logos(error_callback = |lex| {
    format!("Syntax error at {:?}: unrecognised character '{}'", lex.span(), lex.slice())
})]
enum Token {
    #[token("a")]
    A,
    #[token("b")]
    B,
}
```

mysteriouslyseeing:
Closes #444.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -76,6 +76,12 @@ pub trait Logos<'source>: Sized {
     ) -> Lexer<'source, Self> {
         Lexer::with_extras(source, extras)
     }
+
+    /// Create a new error. The default implementation uses `Error::default()`. If you want to make
+    /// your own, use `#[logos(error_callback = ...)]`
+    fn make_error(_lexer: &Lexer<'source, Self>) -> Self::Error {
+        Self::Error::default()
+    }
 }
 
 /// Type that can be returned from a callback, informing the `Lexer`, to skip
--- a/logos-codegen/src/lib.rs
+++ b/logos-codegen/src/lib.rs
@@ -231,6 +231,24 @@ pub fn generate(input: TokenStream) -> TokenStream {
         .take()
         .unwrap_or_else(|| parse_quote!(::logos));
 
+    let make_error_impl = match parser.error_callback.take() {
+        Some(leaf::Callback::Label(label)) => Some(quote! {
+            fn make_error(lex: &#logos_path::Lexer<'s, Self>) -> #error_type {
+                #label(lex)
+            }
+        }),
+        Some(leaf::Callback::Inline(inline)) => {
+            let leaf::InlineCallback { arg, body, .. } = *inline;
+
+            Some(quote! {
+                fn make_error(#arg: &#logos_path::Lexer<'s, Self>) -> #error_type {
+                    #body
+                }
+            })
+        }
+        _ => None,
+    };
+
     let generics = parser.generics();
     let this = quote!(#name #generics);
 
@@ -246,6 +264,8 @@ pub fn generate(input: TokenStream) -> TokenStream {
                 fn lex(lex: &mut #logos_path::Lexer<'s, Self>) {
                     #body
                 }
+
+                #make_error_impl
             }
         }
     };
--- a/logos-codegen/src/parser/mod.rs
+++ b/logos-codegen/src/parser/mod.rs
@@ -29,6 +29,7 @@ pub struct Parser {
     pub skips: Vec<Literal>,
     pub extras: MaybeVoid,
     pub error_type: MaybeVoid,
+    pub error_callback: Option<Callback>,
     pub subpatterns: Subpatterns,
     pub logos_path: Option<TokenStream>,
     types: TypeParams,
@@ -116,6 +117,25 @@ impl Parser {
                     }
                     _ => {
                         parser.err("Expected: #[logos(error = SomeType)]", span);
+                    }
+                }),
+                ("error_callback", |parser, span, value| match value {
+                    NestedValue::Assign(value) => {
+                        let callback = match parser.parse_callback(value) {
+                            Some(callback) => callback,
+                            None => {
+                                parser.err("Not a valid callback", span);
+                                return;
+                            }
+                        };
+                        if let Some(previous) = parser.error_callback.replace(callback) {
+                            parser
+                                .err("Error callback can be defined only once", span)
+                                .err("Previous definition here", previous.span());
+                        }
+                    }
+                    _ => {
+                        parser.err("Expected #[logos(error_callback = ...)]", span);
                     }
                 }),
                 ("extras", |parser, span, value| match value {
--- a/src/lexer.rs
+++ b/src/lexer.rs
@@ -378,11 +378,11 @@ where
         self.token_end = self.source.find_boundary(self.token_end);
         #[cfg(not(feature = "forbid_unsafe"))]
         {
-            self.token = core::mem::ManuallyDrop::new(Some(Err(Token::Error::default())));
+            self.token = core::mem::ManuallyDrop::new(Some(Err(Token::make_error(&self))));
         }
         #[cfg(feature = "forbid_unsafe")]
         {
-            self.token = Some(Err(Token::Error::default()));
+            self.token = Some(Err(Token::make_error(&self)));
         }
     }
 


mysteriouslyseeing:
I'm not sure `error_callback` is the right name for the attribute. Would `error_constructor` be more clear?

jeertmans:
Hi @mysteriouslyseeing, thanks for your PR!

What about using a similar syntax to `skip`:

```
#[logo(error = <ErrorType>)]`
#[logo(error(<ErrorType> [, callback]))]
```
?

E.g., with the new syntax:

```rust
#[logo(error(Error, callback))]
```

book/src/attributes/logos.md
```diff
@@ -34,7 +34,8 @@ The type `ErrorType` can be any type that implements `Clone`, `PartialEq`,
 `Default` and `From<E>` for each callback's error type.
 
 `ErrorType` must implement the `Default` trait because invalid tokens, i.e.,
-literals that do not match any variant, will produce `Err(ErrorType::default())`.
+literals that do not match any variant, will produce `Err(ErrorType::default())`,
+unless you provide a callback with the alternate syntax `#[logos(error(ErrorType, callback = ...))]`
```
jeertmans, line 38: Sorry for the late review. Just one last comment: can you add a bit more details about the callback (signature) and maybe link to the example you wrote? Otherwise, users will not know it is documented :-)

Otherwise, it's all good to be merged!

mysteriouslyseeing:
Just a heads-up that adding a defaulted method to a trait (which this PR does) is [listed as a 'possibly-breaking' change](https://doc.rust-lang.org/cargo/reference/semver.html#trait-new-default-item) in the cargo book's SemVer section.

book/src/attributes/logos.md
mysteriouslyseeing: I actually just changed the custom_error example to use the new error syntax instead of adding a new one, and the example is already linked in the book

jeertmans:
> Just a heads-up that adding a defaulted method to a trait (which this PR does) is [listed as a 'possibly-breaking' change](https://doc.rust-lang.org/cargo/reference/semver.html#trait-new-default-item) in the cargo book's SemVer section.

Good point, meaning that it should be released under a version that either bumps major or minor part.

jeertmans:
According to CI's benchmarks, this PR results in a notable decrease in performance, but I don't really see why: https://github.com/maciejhirsz/logos/actions/runs/12214084915?pr=445.

mysteriouslyseeing:
I ran the benchmarks on my machine and got this:
```
group                                         default_before                         default_changes
-----                                         --------------                         ---------------
count_ok/identifiers                          1.19    618.1±2.13ns  1201.8 MB/sec    1.00    519.2±2.93ns  1431.0 MB/sec
count_ok/keywords_operators_and_punctators    1.13  1711.9±11.62ns  1187.2 MB/sec    1.00  1511.9±26.52ns  1344.2 MB/sec
count_ok/strings                              1.00    409.8±1.47ns  2027.0 MB/sec    1.38    565.8±3.83ns  1468.1 MB/sec
iterate/identifiers                           1.14   590.5±11.93ns  1258.2 MB/sec    1.00    516.1±3.76ns  1439.5 MB/sec
iterate/keywords_operators_and_punctators     1.07  1612.6±16.14ns  1260.3 MB/sec    1.00  1501.0±16.17ns  1354.0 MB/sec
iterate/strings                               1.00    410.9±2.81ns  2021.8 MB/sec    1.38    567.5±4.36ns  1463.6 MB/sec
```
Looks like both strings tests are definitely slower given both the benchmarks. I've got no idea why.

jeertmans:
> I ran the benchmarks on my machine and got this:
> 
> ```
> group                                         default_before                         default_changes
> -----                                         --------------                         ---------------
> count_ok/identifiers                          1.19    618.1±2.13ns  1201.8 MB/sec    1.00    519.2±2.93ns  1431.0 MB/sec
> count_ok/keywords_operators_and_punctators    1.13  1711.9±11.62ns  1187.2 MB/sec    1.00  1511.9±26.52ns  1344.2 MB/sec
> count_ok/strings                              1.00    409.8±1.47ns  2027.0 MB/sec    1.38    565.8±3.83ns  1468.1 MB/sec
> iterate/identifiers                           1.14   590.5±11.93ns  1258.2 MB/sec    1.00    516.1±3.76ns  1439.5 MB/sec
> iterate/keywords_operators_and_punctators     1.07  1612.6±16.14ns  1260.3 MB/sec    1.00  1501.0±16.17ns  1354.0 MB/sec
> iterate/strings                               1.00    410.9±2.81ns  2021.8 MB/sec    1.38    567.5±4.36ns  1463.6 MB/sec
> ```
> 
> Looks like both strings tests are definitely slower given both the benchmarks. I've got no idea why.

Looks like one of the test also increased significantly, that's also weird. I send an e-mail to Logos' author to see if he can enable CodSpeed, which might provide more accurate benchmarks.

mysteriouslyseeing:
The problem is on lines 381 and 385 of src/lexer.rs:
```rust
#[cfg(not(feature = "forbid_unsafe"))]
{
    self.token = core::mem::ManuallyDrop::new(Some(Err(Token::make_error(self))));
}
#[cfg(feature = "forbid_unsafe")]
{
    self.token = Some(Err(Token::make_error(self)));
}
```
Previously, `Token::make_error(self)` was `Token::Error::default()`.

Maybe the problem is that the compiler can't optimise that like it used to? Setting self.token and creating the error both require references to self now - but surely it should be able to see that the reference isn't actually used?

src/lib.rs
```diff
@@ -76,6 +76,12 @@ pub trait Logos<'source>: Sized {
     ) -> Lexer<'source, Self> {
         Lexer::with_extras(source, extras)
     }
+
+    /// Create a new error. The default implementation uses `Error::default()`. If you want to make
+    /// your own, use `#[logos(error_callback = ...)]`
+    fn make_error(_lexer: &mut Lexer<'source, Self>) -> Self::Error {
+        Self::Error::default()
+    }
```
jeertmans, line 84:
```suggestion
    #[inline(always)]
    fn make_error(_lexer: &mut Lexer<'source, Self>) -> Self::Error {
        Self::Error::default()
    }
```

We can probably try this to see if that helps the compiler.

jeertmans (commented):
Thanks for your analysis @mysteriouslyseeing! Let's see if hinting the compiler that it should inline the default implementation can help, see my suggestion.

src/lib.rs
mysteriouslyseeing: Performance is unchanged, unfortunately:
```
group                                         default_before                         default_changes
-----                                         --------------                         ---------------
count_ok/identifiers                          1.20   620.0±18.12ns  1198.2 MB/sec    1.00   518.1±28.66ns  1433.8 MB/sec
count_ok/keywords_operators_and_punctators    1.12  1731.6±95.53ns  1173.6 MB/sec    1.00  1552.2±150.95ns  1309.3 MB/sec
count_ok/strings                              1.00    410.6±6.04ns  2023.2 MB/sec    1.40   574.0±12.94ns  1447.1 MB/sec
iterate/identifiers                           1.15   594.5±30.62ns  1249.7 MB/sec    1.00   518.0±27.87ns  1434.3 MB/sec
iterate/keywords_operators_and_punctators     1.04  1623.9±49.04ns  1251.5 MB/sec    1.00  1559.1±111.50ns  1303.5 MB/sec
iterate/strings                               1.00   412.3±32.70ns  2014.8 MB/sec    1.39   571.8±16.91ns  1452.7 MB/sec
```

jeertmans:
Thanks for your analysis @mysteriouslyseeing! I am not sure to understand your two benchmark results: the first is this PR's changes, and the second is the change between last commit and second-last commit?

mysteriouslyseeing:
> Thanks for your efforts @mysteriouslyseeing! I am not sure to understand your two benchmark results: the first is this PR's changes, and the second is the change between last commit and second-last commit?

Exactly, yes

jeertmans:
Hum, benchmarks results are, *surprising*, as they seem to both indicate performance increase and decrease: https://github.com/maciejhirsz/logos/actions/runs/12279903326/attempts/2#summary-34310133559.

mysteriouslyseeing:
Hmmm, that is strange. Looks like it's mostly increases? Identifiers are worse, but only without `forbid_unsafe`?

jeertmans:
> Hmmm, that is strange. Looks like it's mostly increases? Identifiers are worse, but only without `forbid_unsafe`? Unfortunately I probably can't investigate because my local benchmarks are showing little to no difference in performance.

I'll hold this for one or two more weeks, to see if @maciejhirsz can enable CodSpeed (it should provide better instruments for precise benchmarking). After that, I will try to run benchmark locally and see if it is safe to merge :-) Sorry for the delay!

mysteriouslyseeing:
Just checking in on this?

jeertmans:
> Just checking in on this?

Unfortunately, I did not get any response from @maciejhirsz about enabling CodSpeed, and I would rather not merge any (more) PR that is suspected to decrease performance :-/

mysteriouslyseeing:
Ok, all good. Would you rather leave the issue and PR open or closed?

jeertmans:
> Ok, all good. Would you rather leave the issue and PR open or closed?

Open is fine, and sorry for the inconvenience.

jeertmans:
Hi @mysteriouslyseeing, sorry for the delay in the review!

mysteriouslyseeing:
Yes, I am still available

jeertmans (commented):
Looks great, my only comment is related to https://github.com/maciejhirsz/logos/pull/441#discussion_r1989137874, which I think could make a very nice example to put in the book and/or tests, as this is a very common question we have (i.e., how to parse comments).

jeertmans:
Hi @mysteriouslyseeing, I see that you have been actively working on this for the past few days, thanks! Please ping me once you think it is ready to be reviewed.

mysteriouslyseeing:
Pull [459](https://github.com/maciejhirsz/logos/pull/459) mentions relaxing the error bound to only requiring Default.

The only reason this pull request keep the Default bound is to provide a default implementation of the `make_error` function in the Logos. The reason I went with this initially was to prevent it being a breaking change, but after I looked it up I found it still technically is (as I said [here](https://github.com/maciejhirsz/logos/pull/445#issuecomment-2525198480)).

If the default implementation of `make_error` is removed, the only people affected (requiring a code change) would be those who manually implement the Logos trait, as the default implementation could be added to the code generated by the macro. What do you think? Do you have an idea on how many people manually implement the trait?

Whilst looking into this, I discovered that relaxing trait bounds on associated types is also technically a breaking change, albeit one not listed in the book. The following [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=085f67efde375476715f97b4fdd45f59) example demonstrates this. This would also affect pull 459.

mysteriouslyseeing:
To clarify, even if you provide an error constructor, the error type is still required to implement Default.

jeertmans:
Well, I think this crate really needs more features and bug fixes anyway, so I don't think it's a good idea to stale PRs forever, especially excellent ones like yours! You are right about the potential breaking change induced by this new feature, but I don't mind as long as we document it and use proper versioning to reflect this :-)

Thanks for your contribution!

PR #455, opened 2024-12-19
Fix #[regex] monospace formatting in 4.2 Token and Regex

l1npengtul:
Sorry, just reading through the book and the `#[regex`] bugged me a little...


jeertmans:
Good catch, thanks!

PR #459, opened 2025-01-05
Relax error bound

SarguelUnda:
This relax the bound on logos Error type for more interoperability with error library like anyhow. I could not see where or if those bounds were mandatory. We still need the default trait for constructor but it is ok to use a manual implementation over a wrapper type like

```rust 
struct MyAppLexError(anyhow::Error)
```

It would be harder/misleading to do so for `Clone` and `PartialEq` hence the pull request.


jeertmans:
Hi @SarguelUnda, thanks for your PR!

As you can see, your changes require a increase of the minimal supported Ruest version (MSRV): do you have any idea what is causing this? I guess you can fiund the explaination in the changelog of Rust 1.81? Or maybe is was already broken before, as it seems to be triggered by a dependency we have.

SarguelUnda:
Hello,

from running the command locally it seems to be the home deps from the afl package from the fuzz local crate that trigger this error.

cargo msrv verify -- cargo check --workspace --features debug:

```
Compatibility Check #1: Rust 1.74.0
  [FAIL]   Is incompatible

  ╭───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
  │ error: package `home v0.5.11` cannot be built because it requires rustc 1.81 or newer, while the currently active rustc version is 1.74.0 │
  │ Either upgrade to rustc 1.81 or newer, or use                                                                                             │
  │ cargo update home@0.5.11 --precise ver                                                                                                    │
  │ where `ver` is the latest version of `home` supporting rustc 1.74.0                                                                       │
  │                                                                                                                                           │
  ╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

cargo tree --workspace

```text
logos-fuzz v0.15.0
├── afl v0.15.13
│   ├── libc v0.2.169
│   ├── rustc_version v0.4.1 (*)
│   └── xdg v2.5.2
│   [build-dependencies]
│   └── home v0.5.11
│       └── windows-sys v0.59.0
│           └── windows-targets v0.52.6 (*)
├── arbitrary v1.4.1
└── logos-codegen v0.15.0 (*)
```

sokorototo:
This somewhat tracks, as an error type really only needs `Debug` and `Display`

jeertmans:
Hi @SarguelUnda, sorry for the delay in the review!

interacsion:
I made it work in https://github.com/interacsion/logos/commit/f5aea0cf522dd99a133735f364fb4d9daafcd434, would like me to open a separate PR for it?

jeertmans:
> @jeertmans Would it be possible to also relax the `Default` bound here, since #445 was merged? I find that this requirement is quite frustrating when supplying a custom constructor

Probably, but then we need a way to indicate Logos how to create the default error variant.

robot-rover:
> Just checking: is it by any chance compatible with #491? @robot-rover

Just checked; when I apply the two lines of change all the tests still pass, so it seems yes.

src/lib.rs
DavidCrossman, line 29:
```suggestion
```
This use declaration can be removed.

DavidCrossman:
I've replicated this change on main and everything works fine. I think it would be good to merge this (once the unused use declaration is removed).

DavidCrossman:
I'm in favour of also relaxing the `Default` bound. There are three cases to consider:
1. The user does not supply a custom error type
2. The user supplies a custom error type with a constructor
3. The user supplies a custom error type without a constructor

For (1), the fallback error type is `()`, which implements `Default`, so the bound is irrelevant.
For (2), the user's constructor is called rather than `Default::default`, so the bound isn't needed.
For (3), the `_make_error` function will fallback to `Default::default`. If the custom error type doesn't implement `Default`, this will cause a type error. The user gets an error message suggesting to implement `Default` on the custom error type.

This makes the `Default` bound required only in case (3). The `Logos` derive already uses this technique to require `Into<CustomErrorType>` for each callback's error type.

@jeertmans I can open a separate PR implementing this if you agree. This PR could be merged now; I can fix the unused import in my PR.

PR #475, opened 2025-04-04
Mark read_byte as unsafe and add safety comments

kevincliao:
When forbid_unsafe is enabled, read_byte requires callers to check the next read is still within bounds for it to be safe.

### DIFF
--- a/logos-codegen/src/generator/context.rs
+++ b/logos-codegen/src/generator/context.rs
@@ -59,20 +59,22 @@ impl Context {
         self.available.saturating_sub(self.at)
     }
 
+    /// Reads the next byte without checking bounds.
+    ///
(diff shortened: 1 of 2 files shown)


jeertmans:
Hi @kevincliao, sorry for the late review! I think it makes sense to make the function unsafe, but then the trait's definition changes depending on the safety flag, which I am not sure is 100% intuitive.

I think it's not a breaking as `Context` does not seem to be part of the public API. What do you think?

kevincliao:
Update: Ah nevermind my response about the performance part. Reading the code again, this was the behavior before my PR, so this shouldn't cause any performance regression.

jeertmans:
LGTM, thanks!

Issue #395, opened 2024-06-03
feat(lib): add mermaid diagram debug output + create issue template


jeertmans:
Currently, the `debug` features outputs a diagram that it not very easy to visualize.

As demonstrated in #393, GitHub supports mermaid diagrams and I think this would be an excellent feature to add a `debug` feature that outputs the code for the corresponding mermaid diagram.

Also, this would make sense to create an issue template that asks for users to include to mermaid in the issues, so we can easily visualize what could be wrong!

- [ ] Implement some `debug-mermaid` features;
- [ ] Create an issue template that asks to input the mermaid diagram (when relevant).

PR #477, opened 2025-05-13
Graph generation

### LINKED ISSUE #122: Graph visualization for debugging
A good suggestion [came up on reddit](https://www.reddit.com/r/rust/comments/g3pfaw/logos_011_is_out_iterators_callbacks_and_stateful/fnu70z0/). Could add an extra flag into the top `#[logos]` attribute that points to a path to a dot file that can be used with graphviz.

Should be behind a feature flag on logos-derive so it doesn't increase compile times for people who don't use it.

Shadowcat650:
Closes #122 & #395

Enable the `debug` feature and add this attribute to your enum:
```rust
 #[logos(export_dir = "path/to/export/graphs")] 
```

This will generate a dot graph and mermaid flowchart in that directory.

I noticed while typing the export_dir path, my IDE re-runs the macro every keystroke which creates multiple unwanted directories.

Maybe it would be a good idea to instead create a method on the Token struct that returns the graph strings? This would increase the binary size, but may be worth it if this is a common issue.
```rust
fn dot_str() -> &'static str;
fn mmd_str() -> &'static str;
```


Shadowcat650:
This could also be easily extended to generate mermaid graphs, which was mentioned in #395.

The  `graphviz` feature could be renamed to something like `graph` or graph generation could just be gated behind `debug`.

jeertmans:
Hi @Shadowcat650, thanks for your PR, and sorry for the delay in reviewing it!

> I noticed while typing the dot path, my IDE re-runs the macro multiple times, which creates multiple dot files.

Are you sure this is your IDE running it multiple times? You can easily verify that by cleaning the build files, and running build commands from the terminal (also, raise an error if the `.dot` file already exists, so you know the function is called multiple times).

> This could also be easily extended to generate mermaid graphs, which was mentioned in #395.
> 
> The `graphviz` feature could be renamed to something like `graph` or graph generation could just be gated behind `debug`.

Yes, that sounds great! I think generating a Mermaid diagram during compilation time is probably the best solution (because otherwise we need to expose the inner graph to the user, so it can call `debug` on it, which seems quite complex to do).

Shadowcat650:
Hello @jeertmans!

I decided to gate graph generation behind the `debug` feature. 

It will generate a dot graph and a mermaid graph in the specified directory.
```rust
 #[logos(export_dir = "path/to/export/graphs")
```

Shadowcat650:
> Are you sure this is your IDE running it multiple times? 

Yes. When my IDE detects a change to the macro, it re-runs the macro analyze the output.

book/src/debugging.md
jeertmans, line 1: Great, thanks for the example! You could probably include the full code that was used to generate such graph (instead of writing `...`)?

jeertmans:
I see that your PR is still in draft, is there anything that still needs to be implemented?

jeertmans (commented):
Hi @Shadowcat650, see my review comments and let me know what you think of them :-)

jeertmans:
Thanks for addressing my comments! Looks like a few lines are uncovered (https://github.com/maciejhirsz/logos/pull/477#issuecomment-2878775290): are all trait's methods needed? I think testing for compile errors is not required for all possible errors, but some may be nice (otherwise, this can be left for a future contribution too).

Shadowcat650:
> are all trait's methods needed?

Yes, they weren't covered by the original tests, but the new ones cover them.

jeertmans:
Thanks for your excellent work and contribution @Shadowcat650! We can probably always improve the tool, but let's merge this, as it already provides great debug utilities!

Issue #420, opened 2024-09-13
Error with overlapping token definitions


ccleve:
I'm getting a strange error when a regex could match the prefix of another regex. Maybe. I just don't know what the problem is. Here's a simplified case:

```

#[derive(Logos, Debug, PartialEq)]
#[logos(skip r".|[\r\n]")] // skip everything not recognized
pub enum LogosToken {
    // any letter except capital Z
    #[regex(r"[a-zA-Y]+", priority = 3)]
    WordExceptZ,

    // any number
    #[regex(r"[0-9]+", priority = 3)]
    Number,

    /*
    This expression is:
    (letter or number)* [Z] (letter or number)*
    In other words, a token with any number of letters or numbers,
    including at least one capital Z.
     */
    #[regex(r"[a-zA-Z0-9]*[Z][a-zA-Z0-9]*", priority = 3)]
    TermWithZ,
}

#[pg_extern]
fn test_logos() {
    let mut lex = LogosToken::lexer("hello 42world fooZfoo");
    while let Some(result) = lex.next() {
        let slice = lex.slice();
        println!("{:?} {:?}", slice, result);
    }
}

```

This generates:

```
"hello" Ok(WordExceptZ)
"42world" Err(())
"fooZfoo" Ok(TermWithZ)

```

If I replace the regex over `TermWithZ` with `#[regex(r"Z", priority = 3)]`, I get:

```
"hello" Ok(WordExceptZ)
"42" Ok(Number)
"world" Ok(WordExceptZ)
"foo" Ok(WordExceptZ)
"Z" Ok(TermWithZ)
"foo" Ok(WordExceptZ)

```
The "42world" is getting recognized correctly as a number and word.

What I don't understand is, why does the first TermWithZ regex mess up the recognition of "42world"? It doesn't contain a Z, so TermWithZ should ignore it completely and let the first two variants do their job.

ccleve:
After looking at this a bit more, I'm guessing I'm running into the "no backtracking" limitation.

Any workaround suggestions are welcome.

ccleve:
@jeertmans 
{
    1: ::<skip> (<skip>),
    2: [80-BF] ⇒ 1,
    3: [A0-BF][80-BF] ⇒ 1,
    4: [80-BF][80-BF] ⇒ 1,
    5: [80-9F][80-BF] ⇒ 1,
    6: [90-BF][80-BF][80-BF] ⇒ 1,
[...]

PR #482, opened 2025-06-01
updating Context backtrack for new leaf misses.

martin-juhlin:
when matching similary pattern, this will update the Content::backtrack to incoming nodes miss value, if the miss is pointing to a leaf node, meaning if somehting fail later on, it can go expected value.

adding some new test cases to verify that the new logic is working.

### DIFF
--- a/tests/tests/issue-279.rs
+++ b/tests/tests/issue-279.rs
@@ -1,6 +1,5 @@
 use logos::Logos;
 
-
 #[derive(Logos, Debug)]
 pub enum Token {
     #[token(r"\")]
@@ -16,7 +15,6 @@ pub enum Token {
     #[regex(r"\\[a-zA-Z]+")]
     MacroName,
 }
-
 
 macro_rules! assert_token_positions {
 ($source:expr, $token:pat, $($pos:expr),+ $(,)?) => {


jeertmans:
Hi @Shadowcat650, would you mind rebasing your branch to be in sync with `master`? I just fixed the issue that was blocking tests from running :-)

jeertmans:
Might also fix #265 and #420, tbd

LunarLambda:
Is there anything that's blocking this PR from being merged?

jeertmans:
Not really blocking, I was just hoping we could invest some more time and test if other issues can be closed thanks to this fix.

I would really appreciate it if someone could, for each of the [bugs](https://github.com/maciejhirsz/logos/issues?q=is%3Aissue%20state%3Aopen%20label%3Abug), check if that bug can solved by this PR and, if so, write a test that shows that the bug is now fixed. Similar to what is done here, but only for 3 issues.

LunarLambda:
I see. What about publishing a pre-release version with this merged so we can ask the authors of said issues if they can update to the pre-release and if that fixes things / changes the behaviour? Might speed up the process.

jeertmans:
> I see. What about publishing a pre-release version with this merged so we can ask the authors of said issues if they can update to the pre-release and if that fixes things / changes the behaviour? Might speed up the process.

Hmm I don't have much time to spend on this, and not enough to, so I fear I would just forget about it or postpone this. I would prefer addressing this issue now ^^'

PR #489, opened 2025-06-23
Fix #488

### LINKED ISSUE #488: Already checked len; qed
## MVCE

```rust
#[derive(Logos)]
enum Token {
    #[regex(r"\d+")]
    Integer(),
}
```

## Error

```
error: proc-macro derive panicked
 --> src/main.rs:7:10
  |
7 | #[derive(Logos)]
  |          ^^^^^
  |
  = help: message: Already checked len; qed
```

## Investigation

I think the thought error is here:

https://github.com/maciejhirsz/logos/blob/8b1baf9f3800606ac6fca12f362a3a38bf2851ff/logos-codegen/src/lib.rs#L100-L114

- Yeah, the length is checked (it's 0)
- Yeah, the error case is hit
- However, since there's no kind of flow control after the error emit, the guard check for the length just falls through to the main code after being done → causing a panic

### Suggestion

```diff
                if fields.unnamed.len() != 1 {
                    parser.err(
                        format!(
                            "Logos currently only supports variants with one field, found {}",
                            fields.unnamed.len(),
                        ),
                        fields.span(),
                    );
+                   continue;
                }

                let ty = &mut fields
                    .unnamed
                    .first_mut()
                    .expect("Already checked len; qed")
                    .ty;
                let ty = parser.get_type(ty);

```

All other error checks of this kind do have a corresponding `continue` right afterwards, as far as I skimmed over it. Looks like a one-off.

MultisampledNight:
First adds 2 test snapshots (tuple variants with {0,2} items), then fixes the panic for the former. The last commit is possibly more controversial: It removes the `.expect` call that caused the panic in the first place (but not the bug!) by instead using pattern matching to check for the length while also getting the value.

Should not be a breaking change, only a cosmetic one (the error message gets more comprehensible).

### DIFF
--- a/logos-codegen/src/lib.rs
+++ b/logos-codegen/src/lib.rs
@@ -97,7 +97,7 @@ pub fn generate(input: TokenStream) -> TokenStream {
         let field = match &mut variant.fields {
             Fields::Unit => MaybeVoid::Void,
             Fields::Unnamed(fields) => {
-                if fields.unnamed.len() != 1 {
+                let [ref mut field] = *fields.unnamed.iter_mut().collect::<Box<[_]>>() else {
                     parser.err(
                         format!(
                             "Logos currently only supports variants with one field, found {}",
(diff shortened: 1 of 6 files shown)


MultisampledNight:
Ah, looks like I need to add the snapshots for both <=1.82 and >=1.82 separately. The CI tests for 1.74, I only tested for 1.82 locally. Doing the latter atm, it's exactly the same, will throw it here shortly

MultisampledNight:
1. Regarding the codecov, I will add a new snapshot to test the happy path.
2. The clippy lints look like they'd affect other PRs, too. All of them are in the repo as-is. Should I open a new PR so it can be rebased upon?

PR #497, opened 2025-08-20
Use `package.include` in all published crates

42triangles:
I came across the issue that the whole `mdbook` is included in the published crates during a dependency review, but `logos.png` is especially big and doesn't have to be included for everything to build.

This PR makes use of [`package.include`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields) to not package these files.  
The excluded files based on the current version would be: `RELEASE-PROCESS.md`, `book/**`, `examples/**`, `logos.{svg,png}`, `release.toml`, `tests/**` and `logos-{codegen,cli}/tests/**`.

This reduces the size of almost all crates, but especially `logos` (note the change in units):

| crate | full without `package.include` | compressed | full with `package.include` | compressed |
|--------|--------|--------|--------|--------|
| `logos-codegen` | 275.1KiB | 54.8KiB | 211.0KiB | 49.3KiB |
| `logos-derive` | 20.4KiB | 8.0KiB | 20.5KiB | 8.0KiB |
| `logos` | 1.8MiB | 1.6MiB | 59.0KiB | 18.9KiB |
| `logos-cli` | 48.5KiB | 15.5KiB | 41.4KiB | 14.1KiB |

Changes of data used (using the number of downloads of the previous version, as of 2025-8-20):

| crate | per download | number of downloads per month | bandwidth/month
|--------|--------|--------|--------|
| `logos-codegen` | -5.5KiB | 149416 | -802.5MiB |
| `logos-derive` | ±0.0KiB | 149368 | ±0.0KiB |
| `logos` | -1.6MiB | 149522 | -233.6GiB |
| `logos-cli` | -1.4KiB | 90 | -126.0KiB |

Though it should be noted that older versions also see a lot of downloads as well, so if the same holds for future versions, the actual bandwidth usage decrease would be much more pronounced over time.  
Using _all_ versions over the last 90 days, the `logos` crate would end up with around 2.0TiB/month less bandwidth usage.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ readme = "README.md"
 repository = "https://github.com/maciejhirsz/logos"
 rust-version = "1.74.0"
 version = "0.15.1"
+include = ["build.rs", "/src/", "/examples/", "/tests/", "LICENSE-MIT", "LICENSE-APACHE"]
 
 [package]
 name = "logos"
@@ -32,6 +33,7 @@ readme.workspace = true
 repository.workspace = true
 rust-version.workspace = true
 version.workspace = true
+include.workspace = true
 
 [package.metadata.release]
 pre-release-replacements = [
--- a/logos-cli/Cargo.toml
+++ b/logos-cli/Cargo.toml
@@ -27,6 +27,7 @@ readme.workspace = true
 repository.workspace = true
 rust-version.workspace = true
 version.workspace = true
+include.workspace = true
 
 [package.metadata.release]
 shared-version = true
--- a/logos-codegen/Cargo.toml
+++ b/logos-codegen/Cargo.toml
@@ -39,6 +39,7 @@ readme.workspace = true
 repository.workspace = true
 rust-version.workspace = true
 version.workspace = true
+include.workspace = true
 
 [package.metadata.release]
 shared-version = true
--- a/logos-derive/Cargo.toml
+++ b/logos-derive/Cargo.toml
@@ -24,6 +24,7 @@ readme.workspace = true
 repository.workspace = true
 rust-version.workspace = true
 version.workspace = true
+include.workspace = true
 
 [package.metadata.release]
 shared-version = true


Cargo.toml
```diff
@@ -18,6 +18,7 @@ readme = "README.md"
 repository = "https://github.com/maciejhirsz/logos"
 rust-version = "1.74.0"
 version = "0.15.1"
+include = ["build.rs", "/src", "LICENSE-MIT", "LICENSE-APACHE"]
```
jeertmans, line 21:
```suggestion
include = ["build.rs", "src/**", "LICENSE-MIT", "LICENSE-APACHE"]
```

I think it should be `src/**`, per https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields

jeertmans:
> Though the example from https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields also does remind me: With these changes, `cargo publish` will throw warnings about ignoring examples & tests that aren't included in the final package, too, since they are still referenced in the `Cargo.toml`s, ie:
> 
> ```
> warning: ignoring example `brainfuck` as `examples/brainfuck.rs` is not included in the published package
> ```
> 
> but the resulting package does not need them to build, so that's probably fine - forgot to mention that earlier, my bad!

I think it's still very common to include them in the package: I don't know why, but that might be worth it.

Cargo.toml
42triangles: Which do you prefer? (I personally prefer `/src` or `/src/`, but I'm happy to change it to something else too ^^)

42triangles:
With examples & tests added (`"/examples", "/tests"`), it results in this (again, the _main_ difference is `logos.png`):

| crate | full without `package.include` | compressed | full with `package.include` | compressed |
|--------|--------|--------|--------|--------|
| `logos-codegen` | 275.1KiB | 54.8KiB | 211.0KiB ⇒ 275.2KiB | 49.3KiB ⇒ 54.8KiB |
| `logos-derive` | 20.4KiB | 8.0KiB | 20.5KiB ⇒ 20.6KiB | 8.0KiB ⇒ 8.0KiB |
| `logos` | 1.8MiB | 1.6MiB | 59.0KiB ⇒ 105.0KiB | 18.9KiB ⇒ 28.1KiB |
| `logos-cli` | 48.5KiB | 15.5KiB | 41.4KiB ⇒ 48.6KiB | 14.1KiB ⇒ 15.5KiB |

| crate | per download | number of downloads per month | bandwidth/month
|--------|--------|--------|--------|
| `logos-codegen` | ±0.0KiB | 149416 | ±0.0KiB |
| `logos-derive` | ±0.0KiB | 149368 | ±0.0KiB |
| `logos` | -1.6MiB | 149522 | -233.6GiB |
| `logos-cli` | ±0.0KiB | 90 | ±0.0KiB |

jeertmans:
> With examples & tests added (`"/examples", "/tests"`), it results in this (again, the _main_ difference is `logos.png`):
> 
> crate	full without `package.include`	compressed	full with `package.include`	compressed
> `logos-codegen`	275.1KiB	54.8KiB	211.0KiB ⇒ 275.2KiB	49.3KiB ⇒ 54.8KiB
> `logos-derive`	20.4KiB	8.0KiB	20.5KiB ⇒ 20.6KiB	8.0KiB ⇒ 8.0KiB
> `logos`	1.8MiB	1.6MiB	59.0KiB ⇒ 105.0KiB	18.9KiB ⇒ 28.1KiB
> `logos-cli`	48.5KiB	15.5KiB	41.4KiB ⇒ 48.6KiB	14.1KiB ⇒ 15.5KiB
> crate	per download	number of downloads per month	bandwidth/month
> `logos-codegen`	±0.0KiB	149416	±0.0KiB
> `logos-derive`	±0.0KiB	149368	±0.0KiB
> `logos`	-1.6MiB	149522	-233.6GiB
> `logos-cli`	±0.0KiB	90	±0.0KiB
> It would still be 1.5× the size for `logos` though.
> 
> (Also I am very confused why `logos-derive` gains an _additional_ `0.1KiB` with that, but that was the output of `cargo publish --workspace --dry-run`)

Did you check what files were included with `cargo package --list`?

42triangles:
> Did you check what files were included with `cargo package --list`?

Yup, all tests & examples are included as expected (since it's always relative to the `Cargo.toml` it's referenced in, so ie `logos-codegen/tests` is included from `logos-codegen/Cargo.toml`; and `tests` in the root directory of the repository isn't because it's its own crate)

PR #498, opened 2025-08-20
chore(deps): bump MSRV to 1.80 to use `LazyLock` instead

QnnOkabayashi:


QnnOkabayashi:
Ah unfortunately I think you're right under current conditions. May I ask why the project wants to stay 1.74 compliant?

jeertmans:
Hi @QnnOkabayashi, sorry for the late reply, I totally forgot about this.

> Ah unfortunately I think you're right under current conditions. May I ask why the project wants to stay 1.74 compliant?

There is none, except that it is good practice to document the MSRV accurately and (probably less important) make it gradually evolve, only when needed.

Bumping the MSRV from 1.74 to 1.80 could require a lot of toolchain downloads from end users when they bump our package, while the present PR does not bring any new feature. It just rewrites code with standard library instead.

Something I could be ok with is to declare a new build config variable called `rust_1_80`, see:

https://github.com/maciejhirsz/logos/blob/5234d3c74a0b04afcc29843fd14b5b96735286bd/logos-codegen/build.rs#L1-L21

and use this to include the right construct. See, e.g.:

https://github.com/maciejhirsz/logos/blob/5234d3c74a0b04afcc29843fd14b5b96735286bd/logos-codegen/src/generator/leaf.rs#L32-L36

stefnotch (commented):
This also requires bumping the rust-version to 1.80.0 https://github.com/maciejhirsz/logos/blob/63df35b45d2dbcaebd454db8de36d9c46c3477d3/Cargo.toml#L19

jeertmans:
> It would be nice to clear out old dependencies. Yes, lots of people need to download a new chain, but many many more would not need to recompile `lazy_static` over and over and over.
> 
> 1.80 is a year and a half old. I think it's fine in the Rust ecosystem to make people update.

I didn't realize 1.80 was that old, my bad. I think using a 12-month policy for the MSRV is fine, and so we can accept this. I'll update the PR to have the MSRV corrected.

PR #515, opened 2025-11-25
Update unsafe.md

FigmentEngine:
typo in text, will not work in cargo as feature use different char:
> the feature forbid-unsafe may be
should be a underline _ not a hypen -
> the feature forbid_unsafe may be


jeertmans:
Thank you, @FigmentEngine!

PR #523, opened 2025-12-10
Remove duplicate Logos import from doctests in tests/lib.rs

RoyPrinsGH:
While reading through the book's guidelines on contributing, it [suggested](https://logos.maciej.codes/contributing/setup.html) that I run `cargo test --workspace`. This command failed however, due to these doctests importing Logos both from `logos` and `logos_derive`. 

I removed those double imports and now it's all good :)

Before
image
After
image

### DIFF
--- a/tests/src/lib.rs
+++ b/tests/src/lib.rs
@@ -1,5 +1,5 @@
 //! ```compile_fail
-//! use logos::Logos;
+//! use logos_derive::Logos;
 //!
 //! #[derive(Logos)]
 //! enum Token {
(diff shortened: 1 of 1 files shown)


jeertmans:
Hi @RoyPrinsGH, thanks for catching this! We should probably include doctests within the GitHub workflows. Would you mind editing `.github/workflows/rustlib.yml` such that we also cover those tests?

jeertmans:
Actually, the tests are now failing, see [logs](https://github.com/maciejhirsz/logos/actions/runs/20095424466/job/57652705945?pr=523). This is probably because you ran tests with the wrong configuration... I don't know what's best.

RoyPrinsGH:
It seems I was too eager -- the tests now failed due to not finding `Logos` in the... logos package...

I'll look into it

jeertmans:
Makes sense, thanks for your contribution @RoyPrinsGH!

PR #529, opened 2025-12-24
chore: remove unused dependency

tisonkun:
cc @jeertmans 

`pretty_assertions` and `rustc_version` are purely unused.

I'm trying to replace `beef` with std's Cow. I know `beef` is paired with logos with the same author, but it seems we don't use `beef` in the critical path, so perhaps it's worth trading off for less dependency.

Drop the PR and see your feedback.


jeertmans:
Looks great, thanks!

PR #531, opened 2026-01-04
test: use trybuild for compilation tests

tisonkun:
I'm going to restructure the tests and benches further.

But first, let's leverage [`trybuild`](https://github.com/dtolnay/trybuild/tree/master) for more reliable compilation tests.

This follows https://github.com/maciejhirsz/logos/pull/523#issuecomment-3690102258

cc @jeertmans @robot-rover


jeertmans (approved):
Hi @tisonkun, thank you very much for this PR! I have to admit that it looks much better to (1) use `use logos::*` statement inside rather (rather than a sub-crate) and (2) rely on `trybuild`.

robot-rover:
Just tacking onto this PR: it seems like the trybuild tests are really tanking test times. Locally I was able to compile and run all the tests in a second or so, but the trybuild tests take around 15 seconds. That doesn't seem like much but it really adds up when you test for 2 rust versions and the full suite of possible compilation features, ending up with multiple minutes or more... @tisonkun is there a way to speed these up, or any guidance on caching things so they go faster?

PR #533, opened 2026-01-10
ci: coalesce styles

tisonkun:
Comments inline.

cc @jeertmans


tisonkun:
I'll next organize the tests and bences dirs better, and then start to improve the impl code.

Perhaps we can leverage some tech like [`serde_derive`](https://github.com/serde-rs/serde/tree/master/serde_derive) and `serde_derive_internal` to avoid one more dep level when they are actually shared code.

Besides, `Logos` provides several cool utility for debugging but they are not quite easy for discovering (graph printer, cli, etc.)

Before doing anything above, I typically tend to go through the repo setup so that developing in a well-formed structure rather than workaround any if I encount them later.

PR #535, opened 2026-01-18
chore: move benches under the logos crate

tisonkun:
In a following PR, I'll try to restructure the project layout as:

- examples (a crate in workspace, not published, like [this pattern](https://github.com/fast/logforth/tree/main/examples))
- fuzz (a crate outside workspace, not published, as is)
- logos (a crate in workspace, published, current top-level member crate)
- logos-cli (as is)
- logos-codegen (as is)
- logos-derive (as is)
- tests (as is)


tisonkun:
Ah. I see your feeling now.

Previously, I'm afraid one PR for all restructure too large to review. So instead I made small step as this one.

But based on your concern, I'm going to demonstrate the final state at a whole and decide. We can merge each step one by one later.

For pros and cons, I'd always sync development setup and preference with every opensource project before making significant changes because having a solid codebase enable to contribute with confident: same assumption, CI guarded regression tests, common language for development and communication.

It doesn't mean you should follow my preference, But I'd suggest the goods I know and we will reach a consensus to collaborate together.

Issue #537, opened 2026-01-30
`no_std` broken - Needs a release


pinkforest:
- #521 

Could this be released pretty please ? I'm writing mostly `#[no_std]` code with logos.

```
error[E0433]: failed to resolve: use of unresolved module or unlinked crate `std`
 --> h11types/src/generated/h11header_name_value_tokens.rs:5:17
  |
5 | #[derive(Debug, Logos)]
  |                 ^^^^^ use of unresolved module or unlinked crate `std`
  |
```

Could create a crate in the workspace that tests codegen on no_std and then trigger through CI using the thumb target:
https://github.com/dalek-cryptography/curve25519-dalek/blob/main/.github/workflows/workspace.yml#L61

PR #539, opened 2026-01-31
ci: Add `no_std` / `target` + `cfg` dependant integration tests

pinkforest:
Tests:
- https://github.com/maciejhirsz/logos/issues/537

Can not use trybuild (yet)
- https://github.com/dtolnay/trybuild/issues/326

For now need to add no_std target specific test crate.


jeertmans:
Hi @pinkforest, thanks for your PR! Do you know other projects that test this? If so, please share here :-)

I am not a `no_std` user, so I have to admit that I easily overlook this and I agree that adding tests for this feature would be a great addition. However, I am wondering if there isn't a easier (e.g., using the `cargo hack` command) to test that it works?

PR #542, opened 2026-02-20
Allow multiple lifetimes in tokens

### LINKED ISSUE #148: Lexer extras with lifetime parameters
It would be neat if non-`'static` lifetimes could be used in extras.

```rust
#[derive(Logos)]
#[logos(extras = Option<&'a str>)]
pub enum Foo<'a> {
    #[error]
    Error,
    #[token("bar", |lex| lex.extras.unwrap())]
    Bar(&'a str),
}
```

Currently, this fails to compile:

>error[E0261]: use of undeclared lifetime name `'a`
>```
>    |
>10 | #[logos(extras = Option<&'a str>)]
>    |                          ^^ undeclared lifetime
>```

### LINKED ISSUE #454: Allow other lifetimes in tokens
At the moment, I think Logos forbids having more than one lifetime parameter on a token type, and assumes that the lifetime, if present, is that of the input.

This isn't a fair assumption. I actually have things I want to arena-allocate inside the payload of my tokens, so I need to be able to use the lifetime of the arena.

DavidCrossman:
I've add a new `lifetime` attribute which allows the source lifetime to be specified explicitly, following https://github.com/maciejhirsz/logos/issues/454#issuecomment-2532225245.
This allows multiple lifetimes on token types.
Fixes #454, fixes #148

I've also added to the book to explain the new attribute and how it changes the implicit behaviour. Might need a larger example though.

# Behaviour

## Without attribute
The previous behaviour of the derive macro was to rename the user-provided token lifetime to `'s` in generated code and set the lifetimes in the extras type and in concrete type declarations to `'s`. When the `lifetime` attribute is omitted, this behaviour is maintained so the addition should be backwards compatible. Using multiple lifetimes without the attribute reports an error on each lifetime, explaining that the attribute is required.

## With `lifetime = none`
A fresh lifetime name is generated as the source lifetime; the lifetime `'s` is appended with underscores until it is unique (to avoid clashing with user-defined lifetime names like `'s` or `'s_`). All user-provided lifetime bounds plus the source lifetime are copied to each inner item of the `lex` function (such as the `stageN` functions), so they are in scope with unaltered names in callbacks.

## With `lifetime = 'a`
The lifetime `'a` in the parameter list is used as the source lifetime (errors if not found). `'a` is not renamed to `'s`, so it is available in extras and callbacks, allowing https://github.com/maciejhirsz/logos/issues/148#issue-615077667 to compile (after adding the attribute). All user-provided lifetime bounds (which includes the source lifetime) are copied to each inner item, same as `lifetime = none`.

## Multiple `lifetime` attributes
Reports an error on the attribute explaining that the attribute may only be used once. Also emits an error on the previous attribute.

# Codegen changes

I've replaced the `_Lexer` type alias in the generated code with `use #logos_path::Lexer as _Lexer` and updated the codegen graph snapshots accordingly. I did this because using multiple lifetimes would require each lifetime be added to the type alias to make it available in the token type, which would make the alias more verbose than using its definition directly. I would also need to add another parameter to the `Generator` struct to pass the lifetime list with the bounds stripped if using a type alias.

This is technically a breaking change, since `_Lexer` is in scope in callbacks, but I don't think `_Lexer` is intended to be used by the user. I can change `_Lexer` back to a type alias if this is a concern though.


DavidCrossman:
I've added the following tests to cover what is now possible using this feature:
- Multiple lifetimes (including bounds between lifetimes)
- Static lifetimes in concrete type declarations (previously required a type alias)
- Non-source single lifetime in token type
- Non-source lifetimes in lexer extras

I'm not sure how to test code coverage locally, so let me know if more tests are needed. Thanks.

eggyal:
This may be something for a separate PR, but an `extras` type with multiple lifetimes is not being parsed correctly:

```rust
#[derive(Logos)]
#[logos(extras = Foo<'a, 'b>)] //~ ERROR: Invalid nested attribute; expansion includes `type Extras = Foo<'a;`
#[logos(lifetime = 'a)]
enum Token<'a, 'b> {
    Variant(Foo<'a, 'b>),
}
```

Workaround: wrap type in parens so that it is parsed as a single `TokenTree`.

DavidCrossman:
You already figured out the workaround before I finished, but I'll post this comment anyway to mention the warning.

This is an unrelated issue where Logos doesn't parse `,`s correctly in attributes, but you can work around it by wrapping the type in parentheses, e.g. `(Foo<'a, 'b>)`. Unfortunately, Rust thinks these parentheses should be removed and emits a warning. You can silence the warning like so:
```rust
#[allow(unused_parens)]
mod token {
    use super::*;

    #[derive(Logos)]
    #[logos(extras = (Foo<'a, 'b>))]
    #[logos(lifetime = 'a)]
    pub enum Token<'a, 'b> {
        Variant(Foo<'a, 'b>),
    }
}

use token::*;
```
When the type parameters are concrete (`Bar<'static, u64>`), a type alias could be used instead.

jeertmans (approved):
Thanks for your contribution @DavidCrossman! Could you resolve the conflicts (created from your other PR? :D)

DavidCrossman:
I've fixed the merge conflicts. I'm pretty sure that merging this PR or #555 will create merge conflicts in the other, so I'll fix those as soon as a PR is merged. Thanks for reviewing!

DavidCrossman:
Fixed the merge conflicts caused by merging #555.

jeertmans (approved):
Thanks for your PR @DavidCrossman!

PR #551, opened 2026-04-01
Fix: Clamp offset in end_to_boundary to prevent UB

hieuit095:
When end_to_boundary is called with an offset that exceeds the source length, find_boundary would call is_char_boundary with an out-of-bounds index. For str sources, this is undefined behavior in Rust since is_char_boundary only guarantees correct results for indices <= len. This fix clamps the offset to the source length before calling find_boundary, preventing any potential UB and ensuring the error token span remains valid.


hieuit095:
Closing duplicate. Fix is in PR #552.

PR #553, opened 2026-04-07
Use absolute paths in codegen

DavidCrossman:
I've made all paths in the codegen absolute to improve hygiene. This prevents name clashes if the user writes code such as `struct None;`.


logos-cli/tests/data/fmt_output.rs
jeertmans, line 1: I believe you should not format this file

logos-codegen/src/generator/fast_loop.rs
jeertmans, line 1: Where does `_Option::` come from?

jeertmans (commented):
Hi @DavidCrossman, thanks for your PR! I left to small comments :-)

logos-codegen/src/generator/fast_loop.rs
DavidCrossman: > Where does `_Option::` come from?

The `use ... as ...` on [line 412 in `lib.rs`](https://github.com/maciejhirsz/logos/pull/553/changes/BASE..51a3e0471d65585ad308a49ae6b035000116042a#diff-29dccec76bf4720775864e77adcac53efb245f9bb0f7cd300203d3a22df478d6R412). It looks like an alias to shorten usages of `Option`, so I used it instead of `::core::option::Option` directly.

logos-cli/tests/data/fmt_output.rs
DavidCrossman: > I believe you should not format this file

The `data` folder has `input.rs`, `output.rs` and `fmt_output.rs`, so I thought that `logos-cli input.rs` would update `output.rs` and `logos-cli input.rs --format` would update `fmt_output.rs`.
I think that is correct, but `output.rs` and `fmt_output.rs` haven't been updated in over 3 years so they have changed a lot.

I've reverted the changes to the `data` folder for this PR.
#248 indicates that the `data` folder is from an old test before insta was added. Since #491, `logos-cli` has used insta and both `output.rs` and `fmt_output.rs` are unused. I can remove them in a new PR. What do you think?

logos-codegen/src/generator/fast_loop.rs
jeertmans: Good catch! I didn't see it when reviewing

logos-cli/tests/data/fmt_output.rs
jeertmans: You're correct. Yeah, let's do that in another PR

jeertmans:
thanks for you PR @DavidCrossman!

logos-cli/tests/data/fmt_output.rs
DavidCrossman: I've opened #558 to remove the unused test files in `logos-cli`.

PR #557, opened 2026-04-15
impl fmt::Debug on Lexer: use debug_struct

Tpt:
Closer to what `#[derive(Debug)]` would do

Before:
```
{"source": "", "extras": MockExtras { spaces: 0, line_breaks: 0, numbers: 0, byte_size: 0 }}
```

After:
```
Lexer { source: "", extras: MockExtras { spaces: 0, line_breaks: 0, numbers: 0, byte_size: 0 }, .. }
```

### DIFF
--- a/src/lexer.rs
+++ b/src/lexer.rs
@@ -27,10 +27,10 @@ where
     Token::Extras: Debug,
 {
     fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
-        fmt.debug_map()
-            .entry(&"source", &self.source)
-            .entry(&"extras", &self.extras)
-            .finish()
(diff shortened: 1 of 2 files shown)


jeertmans:
Hi @Tpt, thanks for your contribution! Could you maybe include a before/after comparison of the output?

Tpt:
@jeertmans Thanks! I have updated the MR description

jeertmans:
Looks great! Do you think you could add a test somewhere to verify, using `format!(...)` for example, that the output matches this? This is to avoid possible regression in the future :-)

Tpt:
@jeertmans Done!

jeertmans:
Thanks for your contribution @Tpt!

PR #558, opened 2026-04-15
Remove unused test files from `logos-cli`

DavidCrossman:
Some test files in `logos-cli` are unused and can be removed.
- Since insta has been used for testing snapshots in `logos-cli`, the `logos-cli/tests/data/output.rs` and `logos-cli/tests/data/fmt_output.rs` files have become obsolete.
- #491 added `logos-cli/tests/snapshots/tests__codegen.snap` and `logos-cli/tests/snapshots/tests__codegen_format.snap`, but these snapshots became unused in 24f7707 (before the PR was merged). They are old versions of the current snapshots (only renamed) and can be removed.


jeertmans:
Thanks @DavidCrossman!

Issue #554, opened 2026-04-10
Enable partial input support to support sans I/O


Tpt:
To write sans I/O parser it is very convenient to be able to be able to lex buffers containing not the full input data but only a prefix of it. Then, if the lexer is able to return a "I need more data to properly emit the next token", the parser can return a "please provide more input" result to the user, leading to a I/O call to get more data and a new call to the lexer with a larger buffer.

For example if the grammar is only a single token `<[^>]*>` then, on the `<foo` input a possible `lex_partial` method should be able to return "I need more data" but on `<foo>` it should return the `<foo>` token.

I am currently writing these kind of lexers by hand with a trait like
```rust
trait PartialLexer {
    /// recognize the next token, returns `None` if more data is needed and `is_end` is `false` or if `is_end` is `true` and we have finished lexing the buffer
    fn recognize_next_token<'a>(
        &self,
        buffer: &'a [u8],
        is_end: bool
    ) -> Option<Result<Token<'a>, Error>> {
}
```
but would love to use logos for them.

My uneducated guess is that it can be implemented using the existing automata. The "I need more data" result being emitted if, at the last byte of the `buffer` input, the automaton is in a state with non-error transitions (i.e. a larger token might be matched).

Glad to try an MR if you think it's a good idea.

jeertmans:
Hi @Tpt! I think that's a great idea, and I had wanted to implemented something similar for a long time. Do you think you could come up with a solution that works for anything implementing the `Read` trait? Or `BufRead`? (see #159)

Tpt:
@jeertmans Yes! With this primitive you can easily build something working from `Read` and even `AsyncRead`. The code is roughly something like (abusing a possible generator syntax):
```rust
let mut buffer = vec![0; 2048]; // Initial buffer
let mut buffer_cursor = 0; // offset of first non parsed byte
let mut buffer_end = 0; // offset of last byte in the buffer
loop {
    // We fill again the buffer
    // First we shift the remaining bytes
    if buffer_cursor > 0 {
        buffer.copy_within(buffer_cursor..buffer_end, 0);
        buffer_end -= buffer_cursor;
        buffer_cursor = 0;
    }
    // We ensure we can meaningfully read (in case of very large token)
    if buffer.len() > MAX_BUFFER_SIZE {
        return Err(OOM_ERROR);
    }
    buffer.resize(buffer.len() + 2048);
    // We read
    let read = input.read(&mut buffer)?;
    let is_end = read == 0;
[...]

PR #559, opened 2026-04-18
Adds Lexer::new_prefix to lex a prefix of the input buffer

Tpt:
Details:
- Add Lexer::is_prefix boolean to tag this mode
- When is_prefix is true and the lexing reaches the end of input, we return None if there are still possible normal transitions (i.e. if a larger token is possible)

Adds an example that does stream lexing from a file by filling a buffer, lexing it as much as possible, and then filling it again (all in a loop). It might be nice to move some utilities from the example in the main crate but it's imho outside of MVP.

Issue #554


examples/json_reader.rs
```diff
@@ -0,0 +1,176 @@
+//! Variant of the JSON parser example, parsing from [`Read`] and allowing larger than memeory files.
```
jeertmans:
```suggestion
//! Variant of the JSON parser example, using the [`Read`] trait to avoid storing the complete file in memory.
```

I think this is a better phrasing.

examples/json_reader.rs
jeertmans, line 1: Would be great to include this in the book, as suggested in https://github.com/maciejhirsz/logos/issues/554#issuecomment-4286941757.

Tpt: Done! I have added a new page. Thank you for the review

jeertmans: Thanks!

jeertmans:
Hi @Tpt, thanks for addressing my comments!

Before merging, I would like to discuss the name "prefix lexing": I am not sure this is a common name for such behavior, and wonder if "Buffered lexing" wouldn't make more sense? I.e., replace `new_prefix` with `new_buffered` and `is_prefix` with `is_buffered`. What do you think?

Tpt:
> Before merging, I would like to discuss the name "prefix lexing": I am not sure this is a common name for such behavior, and wonder if "Buffered lexing" wouldn't make more sense? I.e., replace `new_prefix` with `new_buffered` and `is_prefix` with `is_buffered`. What do you think?

@jeertmans I am a bit not at ease by "buffered" because if you load the full input into a buffer, and the lex it, it feels very much "buffered" to me but we won't want the new behavior in this case. But I agree "prefix" is not a great name. What about "partial" like `new_partial` or even `new_partial_prefix` to make things loudly clear?

In any case, glad to move to `new_buffered` if you think it's better, this is your call not mine.

jeertmans:
> > Before merging, I would like to discuss the name "prefix lexing": I am not sure this is a common name for such behavior, and wonder if "Buffered lexing" wouldn't make more sense? I.e., replace `new_prefix` with `new_buffered` and `is_prefix` with `is_buffered`. What do you think?
> 
> @jeertmans I am a bit not at ease by "buffered" because if you load the full input into a buffer, and the lex it, it feels very much "buffered" to me but we won't want the new behavior in this case. But I agree "prefix" is not a great name. What about "partial" like `new_partial` or even `new_partial_prefix` to make things loudly clear?
> 
> In any case, glad to move to `new_buffered` if you think it's better, this is your call not mine.

You're correct that "buffered" may not be correct here, I like your `new_partial` suggestion. Can you implement this change?

Tpt:
@jeertmans Sorry. I let it slip. Done: b6dc674940020cb4ff03c10faa45d40ba7bac1c9

book/src/partial.md
```diff
@@ -0,0 +1,26 @@
...
+
+Usage example
+```rust,no_run,no_playground
+while !is_eof { // We run while we still have unread data
+    let mut lexer = Lexer::new_partial(buffer);
+    while let Some(token) = lexer.next() {
+        // Do something with the token
+    }
+    // We add more data to the buffer
+    let extend_buffer_with_more_data(&mut buffer);
```
jeertmans: Thanks @Tpt! One last minor comment: are you sure about the `let` here?

Tpt: Definitely not
```suggestion
    extend_buffer_with_more_data(&mut buffer);
```

jeertmans:
Great, thank you for your contribution @Tpt!

PR #567, opened 2026-06-08
Add `#[automatically_derived]` to codegen

DavidCrossman:
The [`automatically_derived`](https://doc.rust-lang.org/reference/attributes/derive.html#the-automatically_derived-attribute) attribute has no direct effect, but may provide information to tools. Considering that popular crates such as [`serde`](https://github.com/search?q=repo%3Aserde-rs%2Fserde%20automatically_derived&type=code) and [`derive_more`](https://github.com/search?q=repo%3AJelteF%2Fderive_more%20automatically_derived&type=code) use this attribute, adding it to `impl Logos for` seems reasonable.


jeertmans:
Hi @DavidCrossman, good idea! Thanks for your contribution :-)

PR #570, opened 2026-06-19
Relax `Logos::Error` bounds

DavidCrossman:
## Motivation

`Logos::Error` has several trait bounds which are not necessarily required. These extra bounds impose restrictions on the custom error type the user provides. In particular, the `Default` bound may require adding a `#[default] Other` variant to an error enum even if an error callback is specified.

Relaxing associated trait bounds is a **breaking change**; code generic over the `Logos` trait that relies on `Error`'s trait bounds would need to add those bounds explicitly to migrate, though migration should be straightforward.

## Relaxing the `Default` bound

There are three cases to consider:
1. The user does not supply a custom error type
2. The user supplies a custom error type with an error callback
3. The user supplies a custom error type without an error callback

For (1), the fallback error type is `()`, which implements `Default`, so the bound is irrelevant.
For (2), the user's error callback is called rather than `Default::default`, so the bound isn't needed.
For (3), the `_make_error` function will fallback to `Default::default`. If the custom error type doesn't implement `Default`, this will cause a type error. The user gets an error message suggesting to implement `Default` on the custom error type.

This makes the `Default` bound required only if the user does not provide an error callback (case (3)), rather than unconditionally. The `Logos` derive already uses this technique to require `Into<CustomErrorType>` for each callback's error type.

## Examples and Book
I've updated the book to reflect these changes. I have also simplified the examples which use error callbacks.

## Relation to #459
This PR encompasses #459, so it could be merged before this PR (or closed).


DavidCrossman:
I've realised that relaxing the associated type trait bounds is a breaking change and have updated the PR description.

PR #577, opened 2026-08-18
Add morph variants that replace or map the lexer extras

### LINKED ISSUE #561: Feature request: Lexer::morph variants that deal with extras
Currently `Lexer::morph` requires that the extras of the new token type is `From` of the extras of the old token type.

It would be convenient with a variant of that method which replaces the extras with something else:

```rust
    /// Turn this lexer into a lexer for a new token type and map the extras as appropriate.
    ///
    /// The new lexer continues to point at the same span as the current lexer,
    /// and the current token becomes the error token of the new token type.
    pub fn morph_with_extras<Token2>(self, extras: Token2::Extras) -> Lexer<'source, Token2>
    where
        Token2: Logos<'source, Source = Token::Source>,
    {
        Lexer {
            source: self.source,
            extras,
            token_start: self.token_start,
            token_end: self.token_end,
        }
    }

    /// Turn this lexer into a lexer for a new token type with new extras.
    ///
    /// The new lexer continues to point at the same span as the current lexer,
    /// and the current token becomes the error token of the new token type.
    pub fn morph_map_extras<Token2, F>(self, f: F) -> Lexer<'source, Token2>
    where
        Token2: Logos<'source, Source = Token::Source>,
        F: FnOnce(Token::Extras) -> Token2::Extras,
    {
        Lexer {
            source: self.source,
            extras: f(self.extras),
            token_start: self.token_start,
            token_end: self.token_end,
        }
    }
```

ChrisJr404:
Fixes #561.

Currently `Lexer::morph` requires `Token::Extras: Into<Token2::Extras>`, so morphing between two token types whose extras are unrelated is awkward. This adds three variants that give explicit control over the extras during a morph:

- `morph_with_extras(extras)` — morphs and replaces the extras with a supplied value.
- `morph_map_extras(f)` — morphs and maps the current extras into the new ones with a closure, letting you carry accumulated state across the morph.
- `morph_default_extras()` — morphs and resets the extras via `Default` (the `morph_default_extras` variant suggested in the issue).

All three preserve the current span (`token_start`/`token_end`), the `is_prefix` flag, and the "current token becomes the error token of the new type" semantics, matching `morph`.

Design follows the sketch in #561; `morph_default_extras` is implemented in terms of `morph_with_extras` to keep the behavior consistent.

Tests added in `tests/tests/lexer_modes.rs` cover each variant (extras replacement + span preservation, state carried through the closure, and reset-to-default), and pass under both the default and `forbid_unsafe` features.

### DIFF
--- a/.github/workflows/rustlib.yml
+++ b/.github/workflows/rustlib.yml
@@ -11,6 +11,7 @@ jobs:
   tests:
     name: Tests
     strategy:
+      fail-fast: false
       matrix:
         rust:
         - 1.80.0 # current MSRV
--- a/logos-codegen/src/graph/mod.rs
+++ b/logos-codegen/src/graph/mod.rs
@@ -178,7 +178,7 @@ impl fmt::Display for StateData {
         if f.alternate() {
             writeln!(f, " {{")?;
             for (bc, state) in &self.normal {
-                writeln!(f, "  {} => {}", &bc.to_string(), state)?;
+                writeln!(f, "  {bc} => {state}")?;
             }
             if let Some(eoi_state) = &self.eoi {
                 writeln!(f, "  EOI => {eoi_state}")?;
--- a/logos-codegen/src/lib.rs
+++ b/logos-codegen/src/lib.rs
@@ -354,7 +354,7 @@ pub fn generate(input: TokenStream) -> TokenStream {
                     let matching = matching
                         .iter()
                         .filter(|&id| id != leaf_id)
-                        .map(|match_id| format!("  {}", &graph.leaves()[match_id.0]))
+                        .map(|match_id| format!("  {}", graph.leaves()[match_id.0]))
                         .collect::<Vec<_>>()
                         .join("\n");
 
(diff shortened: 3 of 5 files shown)


jeertmans:
Hi @ChrisJr404, thanks for your contribution! Could you please fix the errors raised by the linter? https://github.com/maciejhirsz/logos/actions/runs/32192020215/job/95935518616?pr=577

ChrisJr404:
Thanks, I cleaned up the clippy lints it was flagging (the redundant references in the format args), so clippy and fmt are both green locally now. One heads up: the nightly test job is hitting an unrelated rustfmt ICE that also reproduces on master, it passes fine on stable.

jeertmans:
Hum I see, but to be sure, could you modify the `.github/workflows/rustlib.yml` workflow flow to set fail fast to false, so I can confirm that it works on stable? E.g., [this is how I do it in another project I own](https://github.com/jeertmans/DiffeRT/blob/8a9cc05b2ab158ccccfb65349a0a01f794651929/.github/workflows/test.yml#L63-L65).

ChrisJr404:
Done, set fail-fast to false on the test matrix so the stable job runs independently of the nightly one.

jeertmans:
Thanks @ChrisJr404

PR #579, opened 2026-08-27
Upgrade logos-codegen to syn 3

jakobhellermann:
Can reduce dependencies if you have syn 3 already in the tree.
`syn` 3.0 has a MSRV of `1.71`, which is lower than this crate.


jeertmans:
Hi @jakobhellermann, thanks! What do you mean by
> Can reduce dependencies if you have syn 3 already in the tree. `syn` 3.0 has a MSRV of `1.71`, which is lower than this crate.

?

logos-codegen/tests/snapshots/codegen__error_callback_failure-tailcall.snap
```diff
@@ -19,7 +20,7 @@ impl<'s> ::logos::Logos<'s> for TokenA {
             { ::core::compile_error!("Previous definition here") }
             {
                 ::core::compile_error!(
-                    "expected one of: `for`, parentheses, `fn`, `unsafe`, `extern`, identifier, `::`, `<`, `dyn`, square brackets, `*`, `&`, `!`, `impl`, `_`, lifetime"
+                    "expected one of: `for`, parentheses, `unsafe`, `fn`, `extern`, identifier, `::`, `<`, `dyn`, square brackets, `*`, `&`, `!`, `impl`, `_`, lifetime"
```
jeertmans, line 23: Why did you change the order? The tests seem to fail now

jakobhellermann: Because I ran the test without the change, and got

```
Snapshot file: logos-codegen/tests/snapshots/codegen__error_callback_failure-tailcall.snap
Snapshot: error_callback_failure-tailcall
Source: logos-codegen/tests/codegen.rs:40
────────────────────────────────────────────────────────────────────────────
Expression: formatted
────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬───────────────────────────────────────────────────────────────
   13    13 │             { ::core::compile_error!("Error type can be defined only once") }
   14    14 │             { ::core::compile_error!("Previous definition here") }
   15    15 │             {
   16    16 │                 ::core::compile_error!(
   17       │-                    "expected one of: `for`, parentheses, `fn`, `unsafe`, `extern`, identifier, `::`, `<`, `dyn`, square brackets, `*`, `&`, `!`, `impl`, `_`, lifetime"
         17 │+                    "expected one of: `for`, parentheses, `unsafe`, `fn`, `extern`, identifier, `::`, `<`, `dyn`, square brackets, `*`, `&`, `!`, `impl`, `_`, lifetime"
   18    18 │                 )
   19    19 │             }
   20    20 │             { ::core::compile_error!("Error type can be defined only once") }
   21    21 │             { ::core::compile_error!("Previous definition here") }
────────────┴───────────────────────────────────────────────────────────────
To update snapshots run `cargo insta review`
Stopped on the first failure. Run `cargo insta test` to run all snapshots.
```

This is just a change in syn2 -> syn3 apparently.

It seems like the tests were still failing because I didn't go far enough in updating snapshot test. I only ran `cargo insta test`, running `cargo insta test --workspace` and accepting the change should hopefully fix CI.

jakobhellermann:
> Hi @jakobhellermann, thanks! What do you mean by
> 
> > Can reduce dependencies if you have syn 3 already in the tree. `syn` 3.0 has a MSRV of `1.71`, which is lower than this crate.
> 
> ?

What I mean is that crates using `logos` that also depend on syn 3 transitively through another dependency will have both syn 2 and syn 3 in their `Cargo.lock` crate graph.
When logos updates to syn 3, they will only have to compile syn once.

With this change, logos only depends on syn 3, at least excluding dev dependencies, which can be checked by running
```
cargo tree -i syn@2 -e no-dev
```

---

Of course, crates that previously only depended on syn 2 (through logos and something else) might now have to compile both syn 2 and syn 3, but that should get fixed as the entire ecosystem upgrades syn.

jeertmans:
Makes sense, thanks! LGTM.

PR #582, opened 2026-09-09
Fix Brainfuck example path and forbid_unsafe feature name in docs

toyeshhm:
Documentation-only corrections; no behavioral change.

- The Brainfuck tutorial (`book/src/examples/brainfuck.md`) and the doc comment at the top of `examples/brainfuck.rs` tell readers to run `cargo run --example brainfuck examples/hello_word.bf`. The file in the repo is `examples/hello_world.bf`, so following the tutorial verbatim fails with a file-not-found error. Both places now name the real file.
- `book/src/unsafe.md` refers to a `forbid-unsafe` feature. The feature is `forbid_unsafe` (as the same page says a few lines earlier, and as declared in `Cargo.toml`), so `--features forbid-unsafe` would be rejected by Cargo.
- Same page: "used in a crates using" → "used in crates using".


jeertmans:
Thanks @toyeshhm!

